PR #16, opened 2020-01-22
correct return types

### LINKED ISSUE #15: Return values of some `AbstractAdapter` methods do not reflect the annotated return type
### Bug Report
There is wrong return type in

zend-cache/src/Storage/Adapter/AbstractAdapter.php

    /**
     * Remove an item.
     *
     * @param  string $key
     * @return bool
     * @throws Exception\ExceptionInterface
     *
     * @triggers removeItem.pre(PreEvent)
     * @triggers removeItem.post(PostEvent)
     * @triggers removeItem.exception(ExceptionEvent)
     */
    public function removeItem($key)

#### Expected behavior

There should be **@return mixed**

There is calling internally
return $this->triggerPost(__FUNCTION__, $args, $result);

which is returnng mixed.

ravcz:


Xerkus:
This change looks wrong.  Proper solution would be to typecast event results to bool. That would be a potential BC break.

boesing:
@Xerkus I am not sure if this might be a BC break as the documentation states a different behavior of what is happening.

This PR is only fixing documentation while casting results from events to boolean might lead to a BC break as we would change behavior.

boesing:
@ravcz could you please fix the DCO issue? 

```bash
$ git commit --amend --signoff
$ git push --force-with-lease origin master
```

After this, this PR should be ready for the next release 👍

boesing:
I will close this for now. Changing all the return types is not what we want to achieve here.
I'll create an Issue targeting the 3.0 release so we ignore the results of the `post` event here and instead we return the result from the adapter.

I actually don't see why the post event could change the return value of the adapter just because anything fails in there.

PR #18, opened 2020-02-11
Add phpredis tls support

LeisureLarry:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

- Are you adding something the library currently does not support?
  - Why should it be added?
    phpredis supports tls since version 5.0
  - What will it enable?
    To use tls normalizeServer needs to support a scheme option.
  - How will the code be used?
    'options' => ['server' => 'tls://127.0.0.1:6380']

### DIFF
--- a/src/Storage/Adapter/RedisResourceManager.php
+++ b/src/Storage/Adapter/RedisResourceManager.php
@@ -229,6 +229,10 @@ class RedisResourceManager
             throw new Exception\InvalidArgumentException('Missing required server host');
         }
 
+        if (isset($server['scheme']) && $server['scheme'] === 'tls') {
+            $host = 'tls://' . $host;
+        }
+
         $server = [
             'host'    => $host,
             'port'    => $port,


src/Storage/Adapter/RedisResourceManager.php
```diff
@@ -229,6 +229,10 @@ protected function normalizeServer(&$server)
             throw new Exception\InvalidArgumentException('Missing required server host');
         }
 
+        if (isset($server['scheme']) && $server['scheme'] === 'tls') {
+            $host = 'tls://' . $host;
+        }
```
Xerkus, line 234: Scheme can already be passed as part of `'host'` in array. That makes this  change  only relevant for string `$server`. So move it into conditional where `parse_url()` is used

Xerkus (changes requested):
This looks sensible. PR will also need documentation update and tests for the change.

boesing:
@LeisureLarry Thanks for contributing to laminas-cache!
The next minor version (2.10.0) of laminas-cache will introduce satellite packages for all storage adapters. Thus, I have to close PRs in here which point to adapters.

If you still need support for this feature, please fork [laminas/laminas-cache-storage-adapter-redis](https://github.com/laminas/laminas-cache-storage-adapter-redis) and re-open a PR in the adapter project.

PR #20, opened 2020-04-01
`sec` property doesn't exists in new MongoDB driver

lorenzomar:
This adapter still use `sec` property from old MongoDate to check item expiration resulting in deletion of the item in any case.

### DIFF
--- a/src/Storage/Adapter/ExtMongoDb.php
+++ b/src/Storage/Adapter/ExtMongoDb.php
@@ -145,7 +145,7 @@ class ExtMongoDb extends AbstractAdapter implements FlushableInterface
                 ));
             }
 
-            if ($result['expires']->sec < (new MongoDate())) {
+            if ($result['expires']->toDateTime() < (new MongoDate())->toDateTime()) {
                 $this->internalRemoveItem($normalizedKey);
                 return;
             }


boesing:
Hey @lorenzomar,

sorry for being late on this. First of all, thanks for your contribution. 🎉 

We are currently working on moving our storage adapters to own repositories. If you are interested in the current state, just checkout our projects page [laminas cache satellites](https://github.com/orgs/laminas/projects/10).

However, as the ext-mongodb driver will find its way into its new repository `laminas-cache-storage-adapter-ext-mongodb`, I will keep this PR on hold for now to not conflict with current progress on the transfer to satellites.

After creating the satellite for the `ext-mongodb` adapter, I will ping you in here so you can create PR against that new component.

Please add some unit tests and make sure that we are backwards compatible. Your current change will have impact on users which are on mongodb less than 5.0. 

I do not fully understand this change as I am not using the cache-component with mongodb but with the satellites, there will be new ways of testing the components aswell. Once we have reached a state that its easy to keep track on changes, we can run integration tests with different mongodb versions and thus making sure that there wont be any BC breaks with this change.

I will keep this issue on my list. With creating the mongodb extension, I will add integration and unit tests for mongodb just to make sure that we can handle different versions.

boesing:
@lorenzomar As already stated in the latest comment, the next minor version (2.10.0) of laminas-cache will introduce satellite packages for all storage adapters. Thus, I have to close PRs in here which point to adapters.

If you still need support for this feature, please fork [laminas/laminas-cache-storage-adapter-ext-mongodb](https://github.com/laminas/laminas-cache-storage-adapter-ext-mongodb) and re-open a PR in the adapter project.

PR #21, opened 2020-04-08
Added plugin interfaces and marked `EventsCapableInterface` as deprecated

boesing:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

As already mentioned in zendframework/zend-cache#177, I would like to add some more interfaces to avoid issues when implementing new adapters without the `AbstractAdapter`.

### DIFF
--- a/src/StorageFactory.php
+++ b/src/StorageFactory.php
@@ -84,7 +84,7 @@ abstract class StorageFactory
 
                 trigger_error(sprintf(
                     'Using "%s" to provide plugin capabilities to storage adapters is deprecated as of '
-                    . 'laminas-cache 2.9; please use "%s" instead',
+                    . 'laminas-cache 2.10; please use "%s" instead',
                     EventsCapableInterface::class,
                     PluginAwareInterface::class
                 ), E_USER_DEPRECATED);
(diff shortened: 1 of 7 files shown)


boesing:
Just curious, do we need that special [`Abstractstorage::__destruct`](https://github.com/laminas/laminas-cache/blob/master/src/Storage/Adapter/AbstractAdapter.php#L90) handling somewhere in a trait or sth? I would prefer to avoid this if its not really necessary.

src/StorageFactory.php
```diff
@@ -72,12 +73,21 @@ public static function factory($cfg)
...
+                    throw new Exception\RuntimeException(sprintf(
+                        "The adapter '%s' doesn't implement '%s' and therefore can't handle plugins",
+                        get_class($adapter),
+                        EventsCapableInterface::class
+                    ));
+                }
+
+                trigger_error(sprintf(
+                    'Using "%s" to provide plugin capabilities to storage adapters is deprecated as of '
+                    . 'laminas-cache 2.9; please use "%s" instead',
```
boesing:
```suggestion
                    . 'laminas-cache 2.10; please use "%s" instead',
```

PR #26, opened 2020-07-16
all datatypes for blackhole adapter

JesBei:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

```PHP
...could not be created. Reason: The storage adapter "Zend\Cache\Storage\Adapter\BlackHole" requires a serializer plugin; please see https://docs.zendframework.com/zend-cache/storage/plugin/#quick-start for details on how to attach the plugin to your adapter.
```

```PHP
... could not be created. Reason: The adapter 'Zend\Cache\Storage\Adapter\BlackHole' doesn't implement 'Zend\EventManager\EventsCapableInterface' and therefore can't handle plugins
```

Blackhole adapter now accepts all datatypes and doesn't need a serializer anymore. Otherwise you have to implement the EventSystem but I don't think that's necessary because blackhole doesn't save anything anyway.

### DIFF
--- a/src/Storage/Adapter/BlackHole.php
+++ b/src/Storage/Adapter/BlackHole.php
@@ -355,7 +355,19 @@ class BlackHole implements
         if ($this->capabilities === null) {
             // use default capabilities only
             $this->capabilityMarker = new stdClass();
-            $this->capabilities     = new Capabilities($this, $this->capabilityMarker);
+            $this->capabilities     = new Capabilities($this, $this->capabilityMarker, [
+                'supportedDatatypes' => [
+                    'NULL'     => true,
+                    'boolean'  => true,
+                    'integer'  => true,
+                    'double'   => true,
+                    'string'   => true,
+                    'array'    => true,
+                    'object'   => true,
+                    'resource' => true,
+                    'light'    => true
+                ],
+            ]);
         }
         return $this->capabilities;
     }
--- a/test/Storage/Adapter/BlackHoleTest.php
+++ b/test/Storage/Adapter/BlackHoleTest.php
@@ -169,4 +169,14 @@ class BlackHoleTest extends TestCase
         $this->assertInstanceOf('Laminas\Cache\Storage\TotalSpaceCapableInterface', $this->storage);
         $this->assertSame(0, $this->storage->getTotalSpace());
     }
+
+    public function testSupportedDataTypes()
+    {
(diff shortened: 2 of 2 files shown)


michalbundyra (changes requested):
@JesBei Thanks for your contribution.

In order to accept the PR we need unit test to be added to cover the changes.
Also please revert all unrelated formatting changes and check DCO status. Thanks.

fabiang:
Btw: the error described by @JesBei happens when using the PSR-16 decorator, which leaves the Blackhole adapter unusable together with PSR-16.

JesBei:
> @JesBei Thanks for your contribution.
> 
> In order to accept the PR we need unit test to be added to cover the changes.
> Also please revert all unrelated formatting changes and check DCO status. Thanks.

Added the Unit Test and reverted the formatting.

boesing:
@JesBei Could you please add integration tests for Blackhole aswell?
There are several integration tests in the `tests/Psr` directory, you can just copy & paste one of these and apply changes to use the Blackhole adapter.

As @fabiang states, there is an issue with the PSR decorators. Thus, with those tests, we would have initial failing tests aswell.

I will move those changes to [laminas/laminas-cache-storage-adapter-blackhole](https://github.com/laminas/laminas-cache-storage-adapter-blackhole) in case we are fixing that "issue" in a patch version. If not, these changes will definitely be part of the 2.10 milestone.

fabiang:
> @JesBei Could you please add integration tests for Blackhole aswell?
> There are several integration tests in the `tests/Psr` directory, you can just copy & paste one of these and apply changes to use the Blackhole adapter.

The integration tests does make much sense, since the blackhole adapter doesn't cache anything and as a result the tests that come from https://github.com/php-cache/integration-tests should fail. The only reasonable test could be that the decorators can handle the adapter and imho this could be done in the unit test directly.

boesing:
@JesBei I have to revert my latest statement, that this will be definetely be part of the next minor (2.10.0).
The next minor version (2.10.0) of laminas-cache will introduce satellite packages for all storage adapters. Thus, I have to close PRs in here which point to adapters.

If you still need support for this feature, please fork [laminas/laminas-cache-storage-adapter-blackhole](https://github.com/laminas/laminas-cache-storage-adapter-blackhole) and re-open a PR in the adapter project.

PR #39, opened 2020-07-27
SimpleCacheDecorator does not fully conform with PSR-16 key length requirements

iganev:
SimpleCacheDecorator does not fully conform with PSR-16 key length requirements. 

PSR-16 does not actually prohibit key lengths of more than 64 characters, if the underlying adapter allows it. 

This is causing issues with numerous 3rd party libraries that require PSR-16 cache, when using `laminas-cache`. 

Proposed solution is to acknowledge the value of `getMaxKeyLength()` from the storage adapter capabilities. 

If the adapter reports unknown max key length, the PSR-16 limit of 64 characters applies.

If the adapter reports unlimited key length, no length check is being performed.

Added unit test to confirm proper behavior.

### DIFF
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -347,10 +347,20 @@ class SimpleCacheDecorator implements SimpleCacheInterface
             ));
         }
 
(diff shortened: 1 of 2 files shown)


samsonasik:
it seems needs rebase.

boesing:
Closing this in favor of #112

PR #17, opened 2020-01-24
Add php 7.4 to travis build matrix

Xerkus:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

PR #44, opened 2020-10-18
update: compatible with phpunit9

gennadigennadigennadi:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description
There are 2 things to discuss:
 - the Depraction of ->at
 - and the use of assertAttributeSame

I just removed those calls because for me it looks like we are testing implementation details that we should not have tested at all.

### DIFF
--- a/test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
+++ b/test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
@@ -8,7 +8,9 @@
 
 namespace LaminasTest\Cache\Psr\CacheItemPool;
 
+use InvalidArgumentException;
 use Laminas\Cache\Exception;
+use Laminas\Cache\Psr\CacheItemPool\CacheException;
 use Laminas\Cache\Psr\CacheItemPool\CacheItemPoolDecorator;
 use Laminas\Cache\Storage\Adapter\AbstractAdapter;
 use Laminas\Cache\Storage\Capabilities;
@@ -26,19 +28,18 @@ class CacheItemPoolDecoratorTest extends TestCase
 
     public function testStorageNotFlushableThrowsException()
     {
-        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
         $storage = $this->prophesize(StorageInterface::class);
 
         $capabilities = new Capabilities($storage->reveal(), new stdClass(), $this->defaultCapabilities);
 
         $storage->getCapabilities()->willReturn($capabilities);
 
+        $this->expectException(CacheException::class);
         $this->getAdapter($storage);
     }
 
     public function testStorageNeedsSerializerWillThrowException()
     {
-        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
         $storage = $this->prophesize(StorageInterface::class);
 
         $dataTypes = [
@@ -59,20 +60,23 @@ class CacheItemPoolDecoratorTest extends TestCase
 
         $storage->getCapabilities()->willReturn($capabilities);
 
+        $this->expectException(CacheException::class);
         $this->getAdapter($storage);
     }
 
     public function testStorageFalseStaticTtlThrowsException()
     {
-        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
         $storage = $this->getStorageProphesy(['staticTtl' => false]);
+
+        $this->expectException(CacheException::class);
         $this->getAdapter($storage);
     }
 
     public function testStorageZeroMinTtlThrowsException()
     {
-        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
         $storage = $this->getStorageProphesy(['staticTtl' => true, 'minTtl' => 0]);
+
+        $this->expectException(CacheException::class);
         $this->getAdapter($storage);
     }
 
@@ -92,7 +96,7 @@ class CacheItemPoolDecoratorTest extends TestCase
      */
     public function testGetItemInvalidKeyThrowsException($key)
     {
-        $this->expectException('Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException');
+        $this->expectException(InvalidArgumentException::class);
         $this->getAdapter()->getItem($key);
     }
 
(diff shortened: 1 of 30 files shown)


test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -16,17 +16,17 @@
...
 {
     use MockStorageTrait;
+    use ProphecyTrait;
 
-    /**
-     * @expectedException \Laminas\Cache\Psr\CacheItemPool\CacheException
-     */
     public function testStorageNotFlushableThrowsException()
     {
+        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
```
Ocramius: Let's use `::class` for these 👍

Ocramius: Also, let's put the expectation right above the line that is expected to fail (not at the beginning of the test)

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -36,11 +36,9 @@ public function testStorageNotFlushableThrowsException()
         $this->getAdapter($storage);
     }
 
-    /**
-     * @expectedException \Laminas\Cache\Psr\CacheItemPool\CacheException
-     */
     public function testStorageNeedsSerializerWillThrowException()
     {
+        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
```
Ocramius: Same ad above: `::class` and closer to throwing statement

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -64,20 +62,16 @@ public function testStorageNeedsSerializerWillThrowException()
         $this->getAdapter($storage);
     }
 
-    /**
-     * @expectedException \Laminas\Cache\Psr\CacheItemPool\CacheException
-     */
     public function testStorageFalseStaticTtlThrowsException()
     {
+        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
```
Ocramius: Same ad above: `::class` and closer to throwing statement

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -64,20 +62,16 @@ public function testStorageNeedsSerializerWillThrowException()
...
         $storage = $this->getStorageProphesy(['staticTtl' => false]);
         $this->getAdapter($storage);
     }
 
-    /**
-     * @expectedException \Laminas\Cache\Psr\CacheItemPool\CacheException
-     */
     public function testStorageZeroMinTtlThrowsException()
     {
+        $this->expectException('Laminas\Cache\Psr\CacheItemPool\CacheException');
```
Ocramius: Same ad above: `::class` and closer to throwing statement

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -95,10 +89,10 @@ public function testGetDeferredItem()
 
     /**
      * @dataProvider invalidKeyProvider
-     * @expectedException \Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException
      */
     public function testGetItemInvalidKeyThrowsException($key)
     {
+        $this->expectException('Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException');
```
Ocramius: `::class`

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -112,11 +106,9 @@ public function testGetItemRuntimeExceptionIsMiss()
         $this->assertFalse($item->isHit());
     }
 
-    /**
-     * @expectedException \Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException
-     */
     public function testGetItemInvalidArgumentExceptionRethrown()
     {
+        $this->expectException('Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException');
```
Ocramius: Same ad above: `::class` and closer to throwing statement

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -167,11 +159,9 @@ public function testGetMixedItems()
         $this->assertTrue($items['bar']->isHit());
     }
 
-    /**
-     * @expectedException \Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException
-     */
     public function testGetItemsInvalidKeyThrowsException()
     {
+        $this->expectException('Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException');
```
Ocramius: Same ad above: `::class` and closer to throwing statement

test/Service/StorageAdapterPluginManagerFactoryTest.php
```diff
@@ -14,24 +14,19 @@
...
+
     public function testFactoryReturnsPluginManager()
     {
         $container = $this->prophesize(ContainerInterface::class)->reveal();
         $factory = new StorageAdapterPluginManagerFactory();
 
         $adapters = $factory($container, AdapterPluginManager::class);
         $this->assertInstanceOf(AdapterPluginManager::class, $adapters);
-
-        if (method_exists($adapters, 'configure')) {
```
Ocramius, line 28: Is this completely gone? ,🤔

Ocramius (changes requested):
Mostly some chore work to be done around exception expectations, but overall good improvement! 👍

test/Service/StorageAdapterPluginManagerFactoryTest.php
gennadigennadigennadi: The code here additionally tests whether $factory sets the container to the AdapterPluginManager - but do you think this is necessary? Since the name of the test is testFactoryReturnsPluginManager, I think it's okay to remove this assertion. 

Otherwise I would need to access the protected property to reimplement the previous assertion.

boesing:
I've created #48 which also adds phpunit 9.4 support.
I'll close this in favor of that PR as it will also close #17

PR #47, opened 2020-11-08
qa: remove all storage adapters

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

As v3.0 switches to dependency inversion, `laminas-cache` only requires a package which provides `laminas/laminas-cache-storage-implementation`


boesing:
Closed in favor of #55

PR #48, opened 2020-11-08
Support for PHP 8.0

### LINKED ISSUE #42: PHP 8.0 support

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no

### Description

Closes #42


test/Pattern/TestAsset/FailableCallback.php
```diff
@@ -2,7 +2,7 @@
 
 namespace LaminasTest\Cache\Pattern\TestAsset;
 
-class FailableCallback
+final class FailableCallback
```
Xerkus, line 5: final is not needed in test assets IMO as they can only be used in this package

boesing: Its not needed but wont hurt so we can keep it.
final aaaall the things.

src/PatternPluginManager.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
weierophinney, line 1: Your new files are missing the license docblock.

src/Service/PatternPluginManagerFactory.php
```diff
@@ -10,17 +10,10 @@
 
 use Interop\Container\ContainerInterface;
 use Laminas\Cache\PatternPluginManager;
-use Laminas\ServiceManager\FactoryInterface;
-use Laminas\ServiceManager\ServiceLocatorInterface;
+use Laminas\ServiceManager\Factory\FactoryInterface;
```
weierophinney: We don't really need to implement the `FactoryInterface` at this point.

weierophinney: Considering that most of these plugin managers will be pulled from the container directly via `get()` and not `build()`, I think it might be interesting to allow passing options to the _constructor_ of the factory (which would also require implementing a `__set_state()` method for serialization purposes). This would allow setting global option when configuring the container.

Thoughts?

src/Service/StorageCacheAbstractServiceFactory.php
```diff
@@ -46,17 +45,6 @@ public function canCreate(ContainerInterface $container, $requestedName)
...
-     * @param  ServiceLocatorInterface $serviceLocator
-     * @param  string $name
-     * @param  string $requestedName
-     * @return boolean
-     */
-    public function canCreateServiceWithName(ServiceLocatorInterface $serviceLocator, $name, $requestedName)
-    {
-        return $this->canCreate($serviceLocator, $requestedName);
-    }
-
```
weierophinney, line 59: Love seeing code deletions; these have been a long time coming!

composer.json
```diff
@@ -28,7 +28,7 @@
         }
```
weierophinney, line 28: You will need to remove the "replace" section of this file; it's no longer a drop-in replacement for the ZF pacakges after the changes you're proposing.

weierophinney (changes requested):
I'm really looking forward to seeing this project done, and seeing the changes for PHP 8, allowing you to drop the old eventmanager and servicemanager versions, as well as aliases for the legacy names, is wonderful!

I have a few additional changes for you, regardless...

composer.json
boesing: Time to drop the bridge aswell as its a new major?

src/Service/PatternPluginManagerFactory.php
boesing: Imho, there is no need to pass any options to the `PatternPluginManager`. 

I'd rather create another PR where I remove `FactoryInterface` from those services/plugin managers which do not need a factory implementing that interface.

If a project might want to add custom stuff to the `PatternPluginManager`, they can just overwrite the factory for it instead of having some "magic" in the existing factory.

Created a PR: https://github.com/boesing/laminas-cache/pull/1

.travis.yml
```diff
@@ -8,43 +8,32 @@ env:
...
         - DEPS=lowest
-    - php: 7.1
+    - php: 7.4
       env:
         - DEPS=latest
         - CS_CHECK=true
         - BENCHMARKS=true
         - TEST_COVERAGE=true
-    - php: 7.2
+    - php: nightly
```
weierophinney, line 38:
```suggestion
    - php: 8.0
```

.travis.yml
```diff
@@ -8,43 +8,32 @@ env:
...
+        - COMPOSER_ARGS="$COMPOSER_ARGS --ignore-platform-reqs"
         - DEPS=lowest
-    - php: 7.2
-      env:
-        - DEPS=latest
-    - php: 7.3
-      env:
-        - DEPS=lowest
-    - php: 7.3
+    - php: nightly
```
weierophinney, line 42:
```suggestion
    - php: 8.0
```

src/PatternPluginManager.php
```diff
@@ -0,0 +1,71 @@
...
+        'Object' => Pattern\ObjectCache::class,
+        'output' => Pattern\OutputCache::class,
+        'Output' => Pattern\OutputCache::class,
+
+        // Legacy Zend Framework aliases
+        \Zend\Cache\Pattern\CallbackCache::class => Pattern\CallbackCache::class,
+        \Zend\Cache\Pattern\CaptureCache::class => Pattern\CaptureCache::class,
+        \Zend\Cache\Pattern\ClassCache::class => Pattern\ClassCache::class,
+        \Zend\Cache\Pattern\ObjectCache::class => Pattern\ObjectCache::class,
+        \Zend\Cache\Pattern\OutputCache::class => Pattern\OutputCache::class,
```
weierophinney, line 40: Get rid of these; you're removing aliases like these from other plugin managers, and you've removed the bridge package, because it targets a new major version. Time to get rid of the legacy!

src/Service/PatternPluginManagerFactory.php
weierophinney: One thing to note: removing the reliance on `FactoryInterface` means you can update to typehint on PSR-11 instead of container-interop interfaces... which I think would be a good idea for this major version.

test/Pattern/CallbackCacheTest.php
```diff
@@ -11,14 +11,14 @@
...
 
 /**
- * Test function
- * @see LaminasTest\Cache\Pattern\Foo::bar
+ * @see \LaminasTest\Cache\Pattern\Foo::bar
  */
 function bar()
 {
-    return call_user_func_array(__NAMESPACE__ . '\TestAsset\TestCallbackCache::bar', func_get_args());
+    return call_user_func_array([TestCallbackCache::class, 'bar'], func_get_args());
```
weierophinney:
```suggestion
    $args = func_get_args();
    return TestCallbackCache::bar(...$args);
```

We have the splat operator; let's simplify!

test/Pattern/ClassCacheTest.php
```diff
@@ -99,20 +99,20 @@ protected function _testCall($method, array $args)
         $return = call_user_func_array([$this->_pattern, $method], $args);
```
weierophinney:
```suggestion
        $return = ($this->_pattern)->$method(...$args);
```

Basically, I'd suggest looking for all `call_user_func_array` usage and updating to use the splat operator.

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -9,39 +9,94 @@
...
 
-class CacheItemPoolDecoratorTest extends TestCase
+final class CacheItemPoolDecoratorTest extends TestCase
 {
-    use MockStorageTrait;
+    protected $defaultCapabilities = [
+        'staticTtl' => true,
+        'minTtl' => 1,
+        'supportedDatatypes' => [
+            'NULL'     => true,
```
weierophinney, line 31: Why caps?

weierophinney (changes requested):
Honestly, looks basically ready. My main concerns are:

- I'd love to get this against PSR-11 instead of container-interop, but that may need a later PR. I'm also aware that things like `AbstractFactoryInterface` implementations require it until laminas-servicemanager is updated.
- Get rid of `call_user_func_array` everywhere.

src/Service/PatternPluginManagerFactory.php
boesing: Thats exactly what I did in boesing/laminas-cache#1
I'll merge that into this branch.

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
boesing: This was just re-added. It was like this in the past and TBH, the only person who could answer that question is @marc-mabe as I did not fully understand any part of this package (yet).

marc-mabe: This was just by the fact that `gettype(null)` returns it as upper case

src/PatternPluginManager.php
boesing: https://github.com/laminas/laminas-cache/pull/55

boesing:
@weierophinney 

I've applied most of the requested changes.
I've referenced #55 as I am dropping all zend-related stuff there aswell.

## Regarding PSR-11 

I'd prefer having those factories (like `InvokableFactory`) in a dedicated (non-servicemanager) component such as `laminas-container-factories` or `laminas-container-contracts` (as you already mentioned in this [RFC](https://discourse.laminas.dev/t/rfc-removal-of-container-interop-from-laminas-servicemanager/1608))

I am facing many situations where I'd love to have a simple `InvokableFactory`.
Its easier to use `InvokableFactory::class` than having multiple anonymous functions for that: 
```php
[
    'factories' => [
         Foo::class => static function () => new Foo(),
         Bar::class => static function () => new Bar(),
    ],
]
```

.travis.yml
```diff
@@ -8,44 +8,43 @@ env:
...
       env:
+        - COMPOSER_ARGS="$COMPOSER_ARGS --ignore-platform-reqs"
         - DEPS=latest
-    - php: 7.3
+    - php: nightly
       env:
+        - COMPOSER_ARGS="$COMPOSER_ARGS --ignore-platform-reqs"
         - DEPS=lowest
-    - php: 7.3
+    - php: nightly
```
weierophinney, line 42: Why nightly as well as 8.0?

If we're pinning to `~8.0.0` in the constraint, _nightly_ is future 8.1.0, and thus out-of-scope.

src/Pattern/CallbackCache.php
```diff
@@ -65,9 +65,9 @@ public function call($callback, array $args = [])
 
         try {
             if ($args) {
-                $ret = call_user_func_array($callback, $args);
+                $ret = $callback(...$args);
             } else {
-                $ret = call_user_func($callback);
+                $ret = $callback();
             }
```
weierophinney: I suggest:

```php
$ret = $args ? $callback(...$args) : $callback();
```

weierophinney (approved):
I've left one question, and one suggestion, but they're really trivial; :ship:

.travis.yml
boesing: Exactly, thats why I've added `allow_failures` for that version. 
This will enable us to early see if we might run out of compatibility in *any* PR rather than waiting until the release.
Its just for visibility tho.

PR #55, opened 2020-11-19
Remove all adapters from plugin manager

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Adapters have to register themselves to the `AdapterPluginManager`.
This can be done by using a `DelegatorFactory` from within the adapter repository. This way, 3rd-party adapters can inject themselves into the `laminas-cache` ecosystem.

### DIFF
--- a/src/Storage/AdapterPluginManager.php
+++ b/src/Storage/AdapterPluginManager.php
@@ -3,7 +3,6 @@
 namespace Laminas\Cache\Storage;
 
 use Laminas\ServiceManager\AbstractPluginManager;
-use Laminas\ServiceManager\Factory\InvokableFactory;
 
 /**
  * Plugin manager implementation for cache storage adapters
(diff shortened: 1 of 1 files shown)


src/Storage/AdapterPluginManager.php
```diff
@@ -22,138 +22,6 @@
...
-        Adapter\Memcache::class                 => InvokableFactory::class,
-        Adapter\Memcached::class                => InvokableFactory::class,
-        Adapter\Memory::class                   => InvokableFactory::class,
-        Adapter\MongoDb::class                  => InvokableFactory::class,
-        Adapter\Redis::class                    => InvokableFactory::class,
-        Adapter\Session::class                  => InvokableFactory::class,
-        Adapter\WinCache::class                 => InvokableFactory::class,
-        Adapter\XCache::class                   => InvokableFactory::class,
-        Adapter\ZendServerDisk::class           => InvokableFactory::class,
-        Adapter\ZendServerShm::class            => InvokableFactory::class,
```
weierophinney: Aren't these still needed?

boesing: theyre provided  by the satellites. 
v3 of this package will use dependency inversion and thus required any package which will implement the `laminas/laminas-cache-storage-adapter-implementation`.

the goal would be, that any Adapter registers itself to the pluginmanager

snapshotpl (approved):

PR #57, opened 2020-12-15
feat: introduce `StorageItem::setItems` for `CacheItemPoolDecorator::commit`

### LINKED ISSUE #56: `CacheItemPoolDecorator::commit` could use `StorageInterface::setItems` for items with same TTL
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

As we are using `StorageInterface` internally, we could take advantage of it by using `StorageInterface::setItems` for items which share the same TTL.

This would speed up processing deferred items when using backends which provide appropriate features.

Lets take `Redis` as an example, which offers the `mSet` or the transactional `setex`  logic (as used in [laminas/laminas-cache-storage-adapter-redis](https://github.com/laminas/laminas-cache-storage-adapter-redis/blob/a306d94cf804b509c4da6c2b81ce872f53b2a673/src/Redis.php#L234)).

There are plenty of other implementations which can store multiple values at once. Thus, I think this addition could make some difference.

Feedback welcome.

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes
| RFC           | yes

### Description

Relates to #56

### DIFF
--- a/src/Psr/CacheItemPool/CacheItemPoolDecorator.php
+++ b/src/Psr/CacheItemPool/CacheItemPoolDecorator.php
@@ -20,11 +20,18 @@ use function array_diff;
 use function array_diff_key;
 use function array_flip;
 use function array_keys;
+use function array_map;
+use function array_merge;
+use function array_unique;
+use function array_walk;
+use function current;
 use function get_class;
 use function gettype;
+use function in_array;
 use function is_string;
 use function preg_match;
 use function sprintf;
+use function var_export;
 
 /**
  * Decorate laminas-cache adapters as PSR-6 cache item pools.
@@ -82,9 +89,9 @@ class CacheItemPoolDecorator implements CacheItemPoolInterface
             }
 
             return new CacheItem($key, $value, $isHit);
-        } else {
-            return clone $this->deferred[$key];
         }
+
+        return clone $this->deferred[$key];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -269,14 +244,22 @@ public function saveDeferred(CacheItemInterface $item)
...
+            $groupedByTtl[$itemTtl] = $group;
+        }
 
-        foreach ($this->deferred as &$item) {
-            if (! $this->save($item)) {
-                $notSaved[] = $item;
-            }
+        $notSavedItems = [];
+        foreach ($groupedByTtl as $keyValuePairs) {
+            $itemTtl         = current($keyValuePairs)->getTtl();
```
boesing, line 258: I could use `groupedByTtl` key here aswell.
But as `null` is an invalid array-key, I'd prefer not to have an additional `if` here.

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -212,40 +220,7 @@ public function save(CacheItemInterface $item)
             throw new InvalidArgumentException('$item must be an instance of ' . CacheItem::class);
```
snapshotpl, line 220: Shouldn't be part of `saveMultipleItems`?

boesing: Nah, its checked in all methods so we can avoid "surprises" later on.
The typehint within `saveMultipleItems` already ensures that we only pass `CacheItem` to it.

The thing is, that you can also `saveDeferred` items which are not directly triggering the save process. Having the check only in `saveMultipleItems` would lead to deferred exceptions.

snapshotpl (approved):

PR #59, opened 2021-01-10
qa: add `laminas/laminas-codingstandard` v2

boesing:
Signed-off-by: Maximilian Bösing <2189546+boesing@users.noreply.github.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Applied coding standard v2.


src/Storage/Adapter/KeyListIterator.php
```diff
@@ -101,15 +102,17 @@ public function setMode($mode)
      */
     public function current()
     {
-        if ($this->mode == IteratorInterface::CURRENT_AS_SELF) {
+        if ($this->mode === IteratorInterface::CURRENT_AS_SELF) {
```
Ocramius, line 105: These changes require careful checking

src/Pattern/ObjectCache.php
```diff
@@ -200,10 +212,13 @@ public function __call($method, array $args)
      * is enabled and the property doesn't exist in real. If so it calls __set
      * and removes cached data of previous __get and __isset calls.
      *
+     * @see    http://php.net/manual/language.oop5.overloading.php#language.oop5.overloading.members
+     *
+     * @phpcs:disable Squiz.Commenting.FunctionComment.InvalidReturnVoid
```
boesing, line 217: Had to add this as some upstream project might added fluent interface stuff to the `call` method.

I'll remove this in a dedicated PR for 3.0.

src/Pattern/ObjectCache.php
```diff
@@ -250,9 +267,13 @@ public function __isset($name)
      * is enabled and the property doesn't exist in real. If so it removes
      * previous cached __isset and __get calls.
      *
+     * @see    http://php.net/manual/language.oop5.overloading.php#language.oop5.overloading.members
+     *
+     * @phpcs:disable Squiz.Commenting.FunctionComment.InvalidReturnVoid
```
boesing, line 272: Had to add this as some upstream project might added fluent interface stuff to the `call` method.

I'll remove this in a dedicated PR for 3.0.

src/PatternPluginManager.php
```diff
@@ -1,67 +1,80 @@
...
 
-    /**
-     * @var string
-     */
+    /** @var string */
     protected $instanceOf = Pattern\PatternInterface::class;
 
-    public function build($plugin, array $options = null)
+    /**
+     * @param string $plugin
```
boesing, line 64: Relates to laminas/laminas-coding-standard#57

src/Storage/Adapter/AbstractAdapter.php
```diff
@@ -106,9 +117,10 @@ public function __destruct()
     /**
      * Set options.
```
boesing, line 118: @laminas/technical-steering-committee We definitely need a sniff which checks if the method description is just repeating the methods name.

src/Storage/AdapterPluginManager.php
```diff
@@ -10,6 +10,22 @@
 
 use Laminas\ServiceManager\AbstractPluginManager;
 use Laminas\ServiceManager\Factory\InvokableFactory;
+use Zend\Cache\Storage\Adapter\Apc;
```
boesing, line 13: Gonna remove these in another PR later on.

src/Storage/Plugin/ClearExpiredByFactor.php
```diff
@@ -30,18 +32,18 @@ public function attach(EventManagerInterface $events, $priority = 1)
     /**
      * Clear expired items by factor after writing new item(s)
      *
-     * @param  PostEvent $event
      * @return void
+     * @phpcs:disable Generic.NamingConventions.ConstructorName.OldStyle
```
boesing, line 36: Cannot rename without introducing BC break.

src/Storage/Plugin/OptimizeByFactor.php
```diff
@@ -27,18 +29,18 @@ public function attach(EventManagerInterface $events, $priority = 1)
     /**
      * Optimize by factor on a success _RESULT_
      *
-     * @param  PostEvent $event
      * @return void
+     * @phpcs:disable Generic.NamingConventions.ConstructorName.OldStyle
```
boesing, line 33: Cannot rename without introducing BC break.

src/Storage/PluginManager.php
```diff
@@ -10,6 +10,11 @@
 
 use Laminas\ServiceManager\AbstractPluginManager;
 use Laminas\ServiceManager\Factory\InvokableFactory;
+use Zend\Cache\Storage\Plugin\ClearExpiredByFactor;
```
boesing, line 13: Gonna remove these in another PR later on.

src/Storage/Adapter/KeyListIterator.php
boesing: I did. 
The mode is being casted to integer in `setMode` method and thus supposed to be integer. 
The property is `protected` and could be set to string even tho it states to be integer.
Do we have to keep such things in mind?

src/Storage/Plugin/ClearExpiredByFactor.php
geerteltink: Wait, what, we still have this?

geerteltink (approved):

src/Storage/Plugin/ClearExpiredByFactor.php
boesing: Yes, but its not used as a `__construct`. Its just a real method which is being invoked from the event manager.

src/Storage/Adapter/KeyListIterator.php
boesing: As this will land in a new major release, I keep this in mind when adding strict typehints. Thus, we would be fine.

PR #63, opened 2021-01-26
CallbackCache: simplify calling function/method

marc-mabe:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Simplified calling the underlying function/method of CallbackCache (and ObjectCache and ClassCache)

### DIFF
--- a/src/Pattern/CallbackCache.php
+++ b/src/Pattern/CallbackCache.php
@@ -58,11 +58,7 @@ class CallbackCache extends AbstractPattern
         // TODO: do not cache on errors using [set|restore]_error_handler
 
         try {
-            if ($args) {
-                $ret = call_user_func_array($callback, $args);
-            } else {
-                $ret = call_user_func($callback);
-            }
+            $ret = $callback(...$args);
         } catch (\Exception $e) {
             if ($cacheOutput) {
                 ob_end_flush();
--- a/src/Pattern/ClassCache.php
+++ b/src/Pattern/ClassCache.php
@@ -40,7 +40,6 @@ class ClassCache extends CallbackCache
         $options   = $this->getOptions();
         $classname = $options->getClass();
         $method    = strtolower($method);
-        $callback  = $classname . '::' . $method;
 
         $cache = $options->getCacheByDefault();
         if ($cache) {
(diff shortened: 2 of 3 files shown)


boesing:
LGTM so far.
Thanks for contributing this!
I'll merge this when I've implemented GHA for CI.
For more details, see laminas/technical-steering-committee#61

snapshotpl (approved):

boesing:
Hm, I was wondering if I could backport this to v2 but actually, this is a BC break.
I've implemented GHA already and it turns out that changing the callback to `[$className, $method]` changes its type from `string` to `array` and thus, leads to an error when the `callback key` is generated.

https://github.com/laminas/laminas-cache/pull/63/files#diff-affcf97709fa42da6b51246d0e1d19dfbd80479d3cfbf2c35eecae7358107e6dR63 
https://github.com/laminas/laminas-cache/blob/15e0b44324a1dae500648ca78bd56e8671565661/src/Pattern/ClassCache.php#L99

I will rebase this PR against 3.0.x as it contains GHA so we can see the failing tests in here.

boesing:
Wait what? I've tried to rebase against 2.12.x but that led into failing tests.
Why the heck isn't that the case in 3.0.x - I am confused...

---

Okay, somehow the names of the test classes changed due to some of the changes made back in february. I've fixed that with #103 and rebased against 3.0.x now.

marc-mabe:
@boesing should be fine now

boesing:
Thanks, Marc. 
Looks good. As far as I can see, this wont be a BC break, right?
I'll probably retarget this against 2.12.x if its not considered BC break so we get this out of the door somewhen in the next weeks rather than months.
WDYT?

boesing (approved):

marc-mabe:
I don't see any bc break here. It's just an internal optimisation.

boesing:
GHA have major outage as of https://www.githubstatus.com/incidents/zbpwygxwb3gw

Verified functionality locally on all three PHP versions.
Thanks, @marc-mabe!

PR #76, opened 2021-04-02
PHP 8.0 support for v2

### LINKED ISSUE #71: PHP 8 support for v2.x
While many laminas components seem to support PHP 8 by now, laminas-cache is currently preventing us from upgrading to PHP 8. I tried to require `"laminas/laminas-cache": "^3.0@dev"`, but `composer update` then fails with `laminas/laminas-cache-storage-adapter-apc 1.0.1 requires php ^5.6 || ^7.0`. I only need the redis adapter btw, but laminas-cache requires all adapters.

Is there an ETA for PHP 8 support or should I look for alternatives of laminas-cache?

boesing:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

Add support for PHP 8.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,6 @@
         "psr/simple-cache": "^1.0"
     },
     "require-dev": {
-        "cache/integration-tests": "^0.16",
         "laminas/laminas-coding-standard": "~1.0.0",
         "laminas/laminas-serializer": "^2.6",
         "phpbench/phpbench": "^0.17.1",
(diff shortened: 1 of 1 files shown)


Slamdunk:
@boesing in order to start satellites PHP8 support, starting from `-test` one, I'd need this PR to be tagged, so I can avoid `dev` requirements in the `composer.json`

boesing:
@Slamdunk There are still open TODOs for 2.11.0 and thus, there is no way of tagging it yet.
You might want to start the initial PR with `@dev` as the `laminas-cache` dependency is `require-dev` anyways.
If not (as in the `apc` storage adapter), that is due to the open merge-up from the release bot. I am actually trying to fix `apc` package by providing `COMPOSER_ROOT_VERSION` but I think I need some more time. So in the end, you can start working locally and creating PR when we are finished preparing the packages to execute proper CI and stuff.

PR #78, opened 2021-04-02
fix potential markdown issues

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes

### Description

Fix potential markdown issues as pointed out by the `markdownlint` github action.

### DIFF
--- a/docs/book/storage/adapter.md
+++ b/docs/book/storage/adapter.md
@@ -18,14 +18,6 @@ these are then passed to an options class instance). Alternately, you can pass
 either the options instance or associative array to the
 `Laminas\Cache\StorageFactory::factory` method.
 
-> **Many Methods throw Exceptions**
->
-> Because many caching operations throw an exception on error, you need to catch
-> them. You can do so manually, or you can use the plugin
-> `Laminas\Cache\Storage\Plugin\ExceptionHandler` with `throw_exceptions` set to
-> `false` to automatically catch them. You can also define an
-> `exception_callback` to log exceptions.
-
 ## Quick Start
 
 Caching adapters can either be created from the provided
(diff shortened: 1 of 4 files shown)


docs/book/storage/adapter.md
```diff
@@ -18,7 +18,7 @@ these are then passed to an options class instance). Alternately, you can pass
 either the options instance or associative array to the
 `Laminas\Cache\StorageFactory::factory` method.
 
-> ### Many Methods throw Exceptions
+> **Many Methods throw Exceptions**
```
froschdesign: We use the third order heading in each blockquote, but for this case we have not defined any rules for Markdown linting.

boesing: Ah, as I see, that works. The problem is, that markdownlint does not properly recognize that this is within a blockquote.

I had similar issues with some `fourth` order heading in the same file, but I wasn't aware that the fix would be to just change them to `third` order heading.

E: What would you think of adding a `second` order heading in front of the blockquote here?

boesing: What do you think of https://github.com/laminas/laminas-cache/pull/78/commits/8773fe19ae0ea4cd39c225e311ad6d4bbc4b21d1

froschdesign: Technical correct but looks strange.

Another suggestion based on the content: move the entire blockquote before the headline "Basic Configuration Options".

## Quick Start

```php
// …

$plugin = new Laminas\Cache\Storage\Plugin\ExceptionHandler();
$plugin->getOptions()->setThrowExceptions(false);
$cache->addPlugin($plugin);
```

> ### Many Methods throw Exceptions
>
> Because many caching operations throw an exception on error, you need to catch
> them. You can do so manually, or you can use the plugin
> `Laminas\Cache\Storage\Plugin\ExceptionHandler` with `throw_exceptions` set to
> `false` to automatically catch them. You can also define an
> `exception_callback` to log exceptions.

## Basic Configuration Options

froschdesign (approved):

PR #81, opened 2021-04-02
docs: add example `composer.json` for required cache adapters

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

With this example, it is possible to only require those adapters which are required by the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,66 @@ storage adapters (DB, File, Memcache, etc).
 - File issues at https://github.com/laminas/laminas-cache/issues
 - Documentation is at https://docs.laminas.dev/laminas-cache/
 
+## Avoid unused cache adapters are being installed
+With `laminas-cache` v2.10.0, we introduced satellite packages for all cache adapters.
+
+In case, there is no need for several adapters in your project, you can use composer to ensure these adapters are not being installed. To make this happen, you have to specify a `replace` property within the `composer.json` of your project.
+
+### Example composer.json with only memory adapter being installed
+```json
+{
+    "name": "vendor/project",
+    "description": "",
+    "type": "project",
+    "require": {
+        "laminas/laminas-cache": "^2.10",
+        "laminas/laminas-cache-adapter-storage-memory": "^1.0"
+    },
+    "replace": {
+        "laminas/laminas-cache-storage-adapter-apc": "*",
+        "laminas/laminas-cache-storage-adapter-apcu": "*",
+        "laminas/laminas-cache-storage-adapter-blackhole": "*",
+        "laminas/laminas-cache-storage-adapter-dba": "*",
+        "laminas/laminas-cache-storage-adapter-ext-mongodb": "*",
+        "laminas/laminas-cache-storage-adapter-filesystem": "*",
+        "laminas/laminas-cache-storage-adapter-memcache": "*",
+        "laminas/laminas-cache-storage-adapter-memcached": "*",
+        "laminas/laminas-cache-storage-adapter-mongodb": "*",
+        "laminas/laminas-cache-storage-adapter-redis": "*",
+        "laminas/laminas-cache-storage-adapter-session": "*",
+        "laminas/laminas-cache-storage-adapter-wincache": "*",
+        "laminas/laminas-cache-storage-adapter-xcache": "*",
+        "laminas/laminas-cache-storage-adapter-zend-server": "*"
+    }
+}
+```
+
+> ### Only necessary in 2.10+
+> Starting with 3.0.0, no storage adapter is required by this component and thus, each project has to specify the storage adapters which are required by the project.
+> When migrated to 3.0.0, the `replace` section is not needed anymore.
+
+When using `composer install` on this, only the `laminas/laminas-cache-storage-adapter-memory` is being installed.
+
+```bash
+Loading composer repositories with package information
+Installing dependencies (including require-dev) from lock file
+Package operations: 10 installs, 0 updates, 0 removals
+  - Installing psr/simple-cache (1.0.1): Loading from cache
+  - Installing psr/cache (1.0.1): Loading from cache
+  - Installing laminas/laminas-zendframework-bridge (1.2.0): Loading from cache
+  - Installing laminas/laminas-stdlib (3.3.1): Loading from cache
+  - Installing psr/container (1.1.1): Loading from cache
+  - Installing container-interop/container-interop (1.2.0): Loading from cache
+  - Installing laminas/laminas-servicemanager (3.6.4): Loading from cache
+  - Installing laminas/laminas-eventmanager (3.3.1): Loading from cache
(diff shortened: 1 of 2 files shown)


froschdesign:
@boesing 
_Sorry for the late response!_

These are migration notes, right? I would suggest to create a page in the documentation and add this information there. Additionally, a link to the migration page should be added in the readme file.

boesing:
@froschdesign No, this is just an information on how to configure a current 2.10 based project to not install all unnecessary cache adapters pre-3.0.
I dont think we can provide PHP 8 support to all adapters as most of them are already not supported on PHP 7 either.

So, we dont have to migrate all adapters to PHP 8 at once and let people decide on which adapters they use and to provide PRs only for these adapters.

So in the end, it is forward compatibility to 3.0 aswell as PHP 8 not having to be supported on every adapter for 2.10+.
Thats the idea behind this example.

froschdesign:
> No, this is just an information on how to configure a current 2.10 based project to not install all unnecessary cache adapters pre-3.0.

Then the description with the example should be added to the installation page.

boesing:
@froschdesign Just to be sure I understand this correct, we dont want to provide informations in `README.md` but via docs only? 

I've also seen that this package is not using the `installation` template other packages are using. So yes, I can extend the installation page here, but how to achieve that in packages which are using the installation template?
Shouldn't we use the template in each package?

froschdesign:
@boesing 
> Just to be sure I understand this correct, we dont want to provide informations in `README.md` but via docs only?

Oh no. In the optimum case we should provide the same informations (like intro, installation, etc.) in the readme file (here on GitHub and after the installation in the vendor directory) and in the documentation, without duplicating the content. Because duplicating always leads to forgotten updates on one side.
So far, we don't have a (good) solution for this, so it reduces to some text and links.

> Shouldn't we use the template in each package?

Yes. I would provide a solution that allows custom content to be added, like the Composer configuration.

README.md
```diff
@@ -11,6 +11,66 @@ storage adapters (DB, File, Memcache, etc).
 - File issues at https://github.com/laminas/laminas-cache/issues
 - Documentation is at https://docs.laminas.dev/laminas-cache/
 
+## Avoid unused cache adapters are being installed
+With `laminas-cache` v2.10.0, we introduced satellite packages for all cache adapters.
```
froschdesign:
```suggestion
## Avoid Unused Cache Adapters Are Being Installed

With `laminas-cache` v2.10.0, we introduced satellite packages for all cache adapters.
```

README.md
```diff
@@ -11,6 +11,66 @@ storage adapters (DB, File, Memcache, etc).
...
+
+In case, there is no need for several adapters in your project, you can use composer to ensure these adapters are not being installed. To make this happen, you have to specify a `replace` property within the `composer.json` of your project.
+
+### Example composer.json with only memory adapter being installed
+```json
```
froschdesign:
```suggestion
### Example `composer.json` with Only Memory Adapter Being Installed

```json
```

README.md
```diff
@@ -11,6 +11,66 @@ storage adapters (DB, File, Memcache, etc).
...
+        "laminas/laminas-cache-storage-adapter-session": "*",
+        "laminas/laminas-cache-storage-adapter-wincache": "*",
+        "laminas/laminas-cache-storage-adapter-xcache": "*",
+        "laminas/laminas-cache-storage-adapter-zend-server": "*"
+    }
+}
+```
+
+> ### Only necessary in 2.10+
+> Starting with 3.0.0, no storage adapter is required by this component and thus, each project has to specify the storage adapters which are required by the project.
```
froschdesign:
```suggestion
> ### Only necessary in 2.10+
>
> Starting with 3.0.0, no storage adapter is required by this component and thus, each project has to specify the storage adapters which are required by the project.
```

docs/book/installation.md
```diff
@@ -1,5 +1,63 @@
-**This is only a placeholder.**
+---
+show_file_content: true
+---
 
-The content of this page can be found under:
+## Avoid unused cache adapters are being installed
+With `laminas-cache` v2.10.0, we introduced satellite packages for all cache adapters.
```
froschdesign:
```suggestion
## Avoid Unused Cache Adapters Are Being Installed

With `laminas-cache` v2.10.0, we introduced satellite packages for all cache adapters.
```

docs/book/installation.md
```diff
@@ -1,5 +1,63 @@
...
+        "laminas/laminas-cache-storage-adapter-session": "*",
+        "laminas/laminas-cache-storage-adapter-wincache": "*",
+        "laminas/laminas-cache-storage-adapter-xcache": "*",
+        "laminas/laminas-cache-storage-adapter-zend-server": "*"
+    }
+}
+```
+
+> ### Only necessary in 2.10+
+> Starting with 3.0.0, no storage adapter is required by this component and thus, each project has to specify the storage adapters which are required by the project.
```
froschdesign:
```suggestion
> ### Only necessary in 2.10+
>
> Starting with 3.0.0, no storage adapter is required by this component and thus, each project has to specify the storage adapters which are required by the project.
```

docs/book/installation.md
```diff
@@ -1,5 +1,63 @@
...
-The content of this page can be found under:
+## Avoid unused cache adapters are being installed
+With `laminas-cache` v2.10.0, we introduced satellite packages for all cache adapters.
 
-https://github.com/laminas/documentation-theme/blob/master/theme/pages/installation.html
\ No newline at end of file
+In case, there is no need for several adapters in your project, you can use composer to ensure these adapters are not being installed. To make this happen, you have to specify a `replace` property within the `composer.json` of your project.
+
+### Example composer.json with only memory adapter being installed
+```json
```
froschdesign:
```suggestion
### Example `composer.json` with Only Memory Adapter Being Installed

```json
```

docs/book/installation.md
froschdesign: Move this block quote to the beginning as it breaks the example and all version 3 users can skip this part / information from the beginning.
(after the headline "Avoid Unused Cache Adapters Are Being Installed")

README.md
froschdesign: Move this block quote to the beginning as it breaks the example and all version 3 users can skip this part / information from the beginning.
(after the headline "Avoid Unused Cache Adapters Are Being Installed")

froschdesign (changes requested):
Some small changes are needed:

* For headline we uses the "[Chicago Manual of Style Capitalization Rules](https://capitalizemytitle.com/#ChicagoStyle)"
* [MD022 - Headings should be surrounded by blank lines](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022)

boesing:
Bummer! I sweared to myself that I will keep an eye on these chicago headlines 😂

froschdesign (approved):
🚢

PR #82, opened 2021-04-02
docs: add example `composer.json` to v2.11

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

With this example, it is possible to only require those adapters which are required by the project.
Relates to #81 and is just rebased against 2.11.x to avoid a release of a new patch of 2.10.3 just to provide `README.md` content which is mostly read on github directly rather than within the `vendor` directory.


boesing:
Closing this as #81 will need a release to push the documentation to the docs.
Therefore, a merge-up is being created anyways.

PR #85, opened 2021-04-03
bugfix: remove items returned true even when removal failed

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

As per definition, `StorageInterface::removeItems` may return a non-empty array of keys which were not removed.
PHP Docblock for `StorageInterface::removeItems`: https://github.com/laminas/laminas-cache/blob/8b9991a31c25c2b7a142334ded74f3938b913337/src/Storage/StorageInterface.php#L194

The `CacheItemPoolDecorator` assumed that the deletion was successful even tho the storage said it was not (by return an array with a list of keys which could not be deleted.

### DIFF
--- a/src/Psr/CacheItemPool/CacheItemPoolDecorator.php
+++ b/src/Psr/CacheItemPool/CacheItemPoolDecorator.php
@@ -194,7 +194,7 @@ class CacheItemPoolDecorator implements CacheItemPoolInterface
         $this->deferred = array_diff_key($this->deferred, array_flip($keys));
 
         try {
-            return null !== $this->storage->removeItems($keys);
+            return $this->storage->removeItems($keys) === [];
         } catch (Exception\InvalidArgumentException $e) {
             throw new InvalidArgumentException($e->getMessage(), $e->getCode(), $e);
         } catch (Exception\ExceptionInterface $e) {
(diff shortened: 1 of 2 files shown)


src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -194,7 +194,7 @@ public function deleteItems(array $keys)
         $this->deferred = array_diff_key($this->deferred, array_flip($keys));
 
         try {
-            return null !== $this->storage->removeItems($keys);
+            return $this->storage->removeItems($keys) === [];
```
Slamdunk, line 197: As per https://github.com/laminas/laminas-cache-storage-adapter-apcu/pull/7#issuecomment-831262536 I think this is a BC Break, take a look at:

https://github.com/php-cache/integration-tests/blob/824633c9547dc3c7ac5faeca03dd82939cdeb73c/src/CachePoolTest.php#L303

Missing keys shouldn't result in a `return false` here.

boesing: Then, that is a problem of the underlying storage.
`StorageInterface::removeItems` is clear on this:

https://github.com/laminas/laminas-cache/blob/15e0b44324a1dae500648ca78bd56e8671565661/src/Storage/StorageInterface.php#L196

I've recently found the same problem in `laminas-cache-storage-adapter-redis`.
The thing is, that it only worked due to the fact that the underlying cache storages had invalid return-types.

I am uncertain on how we should handle this, as there are adapters which actually do return an empty array.
`null` is never returned, it could also be `!== true` in that case.

I'll create a new issue based on your comment to see if we should revert this change and instead "fix" it only in v3.0.0 while also fixing all the adapters.

PR #49, opened 2020-11-08
feat: provide new `StorageAdapterFactory`

boesing:
|    Q          |   A
|-------------- | ------
| Documentation      | yes/no
| BC Break      | yes
| New Feature   | yes

### Description

Introducing `StorageAdapterFactory` to mark `StorageFactory` as deprecated.
Using the factory provides more flexibility as dependency injection for the `AdapterPluginManager` and the `PluginManager` is used instead of relying on globally available static properties.

This reduces complexity and provides better dependency injection support (if the `AdapterPluginManager` and/or the `PluginManager` is received from the container instead).

This addresses #3

PR #86, opened 2021-04-03
Backport of the `StorageAdapterFactory` to v2

### LINKED ISSUE #3: [3.0] Mark static factories as deprecated
As requested in #175

---

Originally posted by @boesing at https://github.com/zendframework/zend-cache/pull/176

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

This is a backport of #49 to provide forward compatibility to the 3.0 release of `laminas-cache`

### DIFF
--- a/docs/book/installation.md
+++ b/docs/book/installation.md
@@ -2,6 +2,35 @@
 show_file_content: true
 ---
 
+## Standalone
+
+If this component is used without `laminas-mvc` or `mezzio`, a PSR-11 container to fetch services, adapters, plugins, etc. is needed.
+
+The easiest way would be to use [laminas-config-aggregator](https://docs.laminas.dev/laminas-config-aggregator/) along with [laminas-servicemanager](https://docs.laminas.dev/laminas-servicemanager/).
+
+```php
+use Laminas\Cache\ConfigProvider;
+use Laminas\Cache\Service\StorageAdapterFactoryInterface;
+use Laminas\Cache\Storage\Adapter\Memory;
+use Laminas\ConfigAggregator\ConfigAggregator;
+use Laminas\ServiceManager\ServiceManager;
+
+$config = (new ConfigAggregator([
+    ConfigProvider::class,
+]))->getMergedConfig();
+
+$dependencies = $config['dependencies'];
+
+$container = new ServiceManager($dependencies);
+
+/** @var StorageAdapterFactoryInterface $storageFactory */
+$storageFactory = $container->get(StorageAdapterFactoryInterface::class);
+
+$storage = $storageFactory->create(Memory::class);
+
+$storage->setItem('foo', 'bar');
+```
+
 ## Avoid Unused Cache Adapters Are Being Installed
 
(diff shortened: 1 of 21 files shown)


docs/book/psr16.md
```diff
@@ -15,12 +15,15 @@ operations.
...
-use Laminas\Cache\StorageFactory;
 use Laminas\Cache\Psr\SimpleCache\SimpleCacheDecorator;
+use Laminas\Cache\Service\StorageAdapterFactoryInterface;
+use Psr\Container\ContainerInterface;
 
-$storage = StorageFactory::factory([
-    'adapter' => 'apc',
-]);
+/** @var ContainerInterface $container */
+$container = null; // can be any configured PSR-11 container
```
boesing, line 23: @froschdesign Should I refer to the `Standalone` section to provide a way how the `container` could be "retrieved"?

Maybe by adding something like:

```markdown
> ### Unsure How to Retrieve a Container Instance?
>
> Check out the [installation page](https://docs.laminas.dev/laminas-cache/installation/)
```

In most scenarios I am working with the `laminas-cache`, I have to retrieve a storage from within a factory which already passes the `ContainerInterface`.
But for those who are not working in a MVC or Mezzio project, that must not be the case.

froschdesign: @boesing 
> Should I refer to the `Standalone` section to provide a way how the `container` could be "retrieved"?

Stand-alone is correct because the component can be used outside of laminas-mvc or Mezzio based application. The basic usage and all other examples should follow this fact.

The integration in a laminas-mvc, Mezzio or other application must be described under the section "Application Integration". (laminas-mvc controller plugins and similar packages are clear exceptions)

docs/book/installation.md
```diff
@@ -2,6 +2,35 @@
...
+$container = new ServiceManager($dependencies);
+
+/** @var StorageAdapterFactoryInterface $storageFactory */
+$storageFactory = $container->get(StorageAdapterFactoryInterface::class);
+
+$storage = $storageFactory->create(Memory::class);
+
+$storage->setItem('foo', 'bar');
+```
+
```
froschdesign: A page for the basic usage is missing and I think this section can be added there.

docs/book/psr16.md
boesing: @froschdesign 

Hm, I still dont really know how to provide proper examples for the "config-driven" creations of the storage adapters.

Starting with v3.0+, the config-driven approach wont work without having a PSR-11 container with the new `StorageAdapterFactoryInterface` configured, e.g. Manual instantiation will ofc still work.

How could I address that properly? I also don't get how an "Application Integration" section might look like.

boesing:
@froschdesign I'll go for this now and if we can improve documentation, I think we can do this in an additional PR/Release.

In v3, we have to adjust the `basic-usage` anyways as *any* cache adapter which is used will have its own `ConfigProvider` which then has to be added to the `ConfigAggregator` as well.
In v2, this is not necessary.

PR #108, opened 2021-05-13
Set static variable to NULL

### LINKED ISSUE #107: Static property cannot be unset
>  If a static variable is `unset() `inside of a function, `unset()` destroys the variable only in the context of the rest of a function.  See: [function.unset.php](https://www.php.net/manual/en/function.unset.php)

https://github.com/laminas/laminas-cache/blob/4250fc8611af994066fd4e02ad4af5e1673e5ff1/src/Pattern/ClassCache.php#L149-L160

ghostwriter:
- resolves #107

This way the next time the function is called it will use `null` as the value for that `static::variable` as it was the last value of that `static::variable` before it was supposed  to have be `unset(static::variable);`.

### DIFF
--- a/docs/book/pattern/class-cache.md
+++ b/docs/book/pattern/class-cache.md
@@ -124,7 +124,7 @@ class ClassCache extends CallbackCache
     public function __unset($name)
     {
         $class = $this->getOptions()->getClass();
-        unset($class::$name);
+        $class::$name = null;
     }
 }
 ```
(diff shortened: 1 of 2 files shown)


boesing:
As mentioned in #107, this probably might cause other issues and thus I've created #118 to mark this pattern as deprecated and thus, I'd say lets ignore this until 3.0.

PR #112, opened 2021-05-16
PSR-16 decorator should use maximum key length capability

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no

### Description

Since its very first version, the `PSR-16` decorator only accepted keys up to 64 characters.
The user iganev provided a solution in #39 which based on the maximum key length capability of the storage adapters.

It seems that most of the storage adapters do return proper key lengths ([except redis](https://github.com/laminas/laminas-cache-storage-adapter-redis/issues/9)) and thus, this change seems absolutely legit to me.

So with v2.12.0 I'd introduce this change and with v3.0.0, all adapters returning `-1` as maximum key length will be incompatible with the `PSR-16` decorator.

### DIFF
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -342,19 +342,19 @@ class SimpleCacheDecorator implements SimpleCacheInterface
         }
 
         // get storage adapter capability value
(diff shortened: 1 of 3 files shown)


boesing:
Thanks @iganev!

PR #118, opened 2021-05-24
Mark `ClassCache` pattern as deprecated

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

As recommended by @marc-mabe in #107, we mark the `ClassCache` as deprecated as of v2.12.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,6 +55,7 @@
         "laminas/laminas-cli": "^1.0",
         "laminas/laminas-coding-standard": "~1.0.0",
         "laminas/laminas-config-aggregator": "^1.5",
+        "laminas/laminas-feed": "^2.14",
         "laminas/laminas-serializer": "^2.6",
         "phpbench/phpbench": "^1.0.0-beta2",
         "phpspec/prophecy-phpunit": "^2.0",
--- a/docs/book/pattern/class-cache.md
+++ b/docs/book/pattern/class-cache.md
@@ -1,9 +1,10 @@
 # ClassCache
 
-The `ClassCache` pattern is an extension to the
-[`CallbackCache`](callback-cache.md) pattern. It has the same methods, but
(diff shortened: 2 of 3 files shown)


marc-mabe:
It should be marked as deprecated in documentation as well : https://github.com/laminas/laminas-cache/blob/2.12.x/docs/book/pattern/class-cache.md

marc-mabe:
Probably also trigger a deprecation notice on class loading or initializing ?

boesing:
Ive thought it would be a good thing as well when introducing the new Interfaces for plugins but there, the notice was only triggered when upstream did not implemented the `AbstractStorage`. In this case, it would trigger like always as it is a manual pick.

There were some complaints from @Ocramius about having runtime effects in deprecations and thus I try to avoid these now as I see the point. 
Since the class will only be removed in the next major, I think it will be fine. Updating to a new major should be a process where the migration documentation should be used and thus I have to add this there. 

Actually, there is not even an alternative to this and thus no way of fixing the deprecation message unless someone copys the code and removes the message.

What do you think?

marc-mabe:
Fine for me. At least the current documentation and (upcoming) migration guide should include this deprecation/removal.

boesing:
@froschdesign Is there some "standard" to mark something as deprecated in the documentation?

froschdesign:
@boesing 
We use a blockquote for this. Example:

``````
# ClassCache

> ### Deprecated
>
> The `ClassCache` pattern is deprecated…
``````

The block should be placed directly after the headline. It would be good if an alternative with a link could be added.

boesing:
@froschdesign Sadly, there is no alternative yet. I am thinking about creating a pattern which can be used in combination with PSR-6 to have such cache logic again but that is not yet implemented.

PR #155, opened 2021-08-05
Update SimpleCacheDecorator.php

10n:
// PSR-16 does not allow more then 64 bytes for the key and should not be more than the capabilities of the storage

|    Q          |   A
|-------------- | ------ 
| Bugfix        | yes  

### Description

Fixes an issue with SimpleCacheDecorator (PSR-16) that was forcing all storages to have the capability to have a key for at least 64 bytes, but the PSR-16 specifications indicate that a maximum of 64 bytes should be allowed. Having a property that stores the `maximumKeyLength` based on the storage capability and used to validated each key, would mean that smaller keys should be allowed if their length do not exceed the capability of the storage or the specs.

This also fixes PHP warning generated by the pattern used in `preg_match` to validate the length of each key.

### DIFF
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -448,7 +448,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
             $maximumKeyLength = 64;
         }
 
-        // PSR-16 does not allow more then 64 bytes for the key and should not be more than the capabilities of the storage
+        // PSR-16 does not allow more than 64 bytes for the key and should not be more than the capabilities of the storage
         $this->maximumKeyLength = min($maximumKeyLength, 64);
     }
 }
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -448,14 +448,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
             $maximumKeyLength = 64;
         }
 
-        if ($maximumKeyLength < 64) {
-            throw new SimpleCacheInvalidArgumentException(sprintf(
-                'The storage adapter "%s" does not fulfill the minimum requirements for PSR-16:'
-                .' The maximum key length capability must allow at least 64 characters.',
-                get_class($storage)
-            ));
-        }
-
-        $this->maximumKeyLength = $maximumKeyLength;
+        // PSR-16 does not allow more then 64 bytes for the key and should not be more than the capabilities of the storage
+        $this->maximumKeyLength = min($maximumKeyLength, 64);
     }
 }


boesing:
PSR-16 states that at least 64 characters have to be supported. Thats the minimum, not the maximum.
You can read about this here: https://www.php-fig.org/psr/psr-16/


> Key - A string of at least one character that uniquely identifies a cached item. Implementing libraries MUST support keys consisting of the characters `A-Z`, `a-z`, `0-9`, `_`, and `.` in any order in UTF-8 encoding and a length of up to 64 characters. Implementing libraries **MAY support additional characters and encodings or longer lengths, but MUST support at least that minimum.** Libraries are responsible for their own escaping of key strings as appropriate, but MUST be able to return the original unmodified key string. The following characters are reserved for future extensions and MUST NOT be supported by implementing libraries: `{}()/\@:`

There are also integration tests which checks that:
https://github.com/php-cache/integration-tests/blob/824633c9547dc3c7ac5faeca03dd82939cdeb73c/src/SimpleCacheTest.php#L422
https://github.com/php-cache/integration-tests/blob/824633c9547dc3c7ac5faeca03dd82939cdeb73c/src/CachePoolTest.php#L122

10n:
I believe you are misreading the specification https://www.php-fig.org/psr/psr-16/ . This states:

> Key - A string of at least one character that uniquely identifies a cached item. Implementing libraries MUST support keys consisting of the characters A-Z, a-z, 0-9, _, and . in any order in UTF-8 encoding **and a length of up to 64 characters**.

That means that keys above 64 characters are not accepted and should not be allowed.  

In previous implementation the validation was explicitly throwing an exception if the key was to big. So there was a hardcoded upper limit (exactly how the PSR-16 specifies)/
```
new SimpleCacheInvalidArgumentException(sprintf(
 'Invalid key "%s" provided; key is too long. Must be no more than 64 characters',
 $key
));
```
https://github.com/laminas/laminas-cache/blob/775cf8815cfd5a654f9b7b341ebaf28e72aa457d/src/Psr/SimpleCache/SimpleCacheDecorator.php#L350

The current implementation validates the capabilities of the storage (accepts only those storages that accept keys with at least 64 characters).

10n:
Nevermind, I was the one mistaken.

boesing:
Just to catch up as I guess that this will probably come up again:

> That means that keys above 64 characters are not accepted and should not be allowed.

Actually, thats not the case. Even tho, the wording is very misleading, I think that the following should be verified:
1. the underlying storage implementation **MUST** support up to 64 characters
2. if the underlying storage can support more than 64 characters, thats fine. But it **MUST** support at least 64 characters.

Thats how I understand that part right after your quoted part of the PSR. And when checking the implementation tests provided by `php-cache`, I think thats actually correct.

10n:
Now I understand, and thank you for taking time to explain.

After reading it again ( for the fifth time I think) I can understand now that this Decorator should be able to replace any other library that implements the same protocol/standard. 

That means that it should satisfy those minimum requirements accepted characters and maximum possible length.

Thank you once again.

PR #169, opened 2021-10-24
Widening return type hint of SimpleCacheDecorator::translateThrowable method

mariojrrc:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Tell us about why this change is necessary:
- Are you fixing a bug or providing a failing unit test to demonstrate a bug?
no
- How do you reproduce it?
Set up a Filesystem adapter and try to call has/get/set methods of Laminas\Cache\Psr\SimpleCache\SimpleCacheDecorator passing an invalid key such as '127.0.0.1'

- What did you expect to happen?
It should throws the Laminas\Cache\Psr\SimpleCache\SimpleCacheInvalidArgumentException with message "The key '127.0.0.1' doesn't match against pattern '/^[a-z0-9_\\+\\-]*$/Di'"

- What actually happened?
It throws a TypeError exception with the following message: 

"Return value of Laminas\\Cache\\Psr\\SimpleCache\\SimpleCacheDecorator::translateThrowable() must be an instance of Laminas\\Cache\\Psr\\SimpleCache\\SimpleCacheException, instance of Laminas\\Cache\\Psr\\SimpleCache\\SimpleCacheInvalidArgumentException returned"

- TARGET THE 2.13.x BRANCH


src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -291,7 +291,11 @@ public function has($key)
         }
     }
 
-    private static function translateThrowable(Throwable $throwable): SimpleCacheException
+    /**
+     * @param Throwable $throwable
+     * @return SimpleCacheException|SimpleCacheInvalidArgumentException
+     */
+    private static function translateThrowable(Throwable $throwable)
```
boesing: Instead of removing the type-hint, we should open-up the type-hint to `Psr\SimpleCache\CacheException`.
Thats the interface which has to get implemented by all exceptions which are thrown from any PSR-16 implementation.

boesing (changes requested):
Thanks for pointing out that issue.
Could you please also create a failing test case? There should be no need of the filesystem adapter to trigger this method.

src/Psr/SimpleCache/SimpleCacheDecorator.php
mariojrrc: You're right. I'm gonna update that.

mariojrrc: @boesing before my modificacion a simple failing test looks like this.

```
public function testWillThrowExceptionWhenUsingFilesystemAdapterAndUseInvalidCacheKey(): void
    {
        $storage = StorageFactory::factory([
            'adapter' => 'filesystem',
            'plugins' => ['Serializer'],
        ]);

        $this->assertInstanceOf(Filesystem::class, $storage);

        $decorator = new SimpleCacheDecorator($storage);

        $this->expectException(InvalidArgumentException::class);
        $decorator->get('127.0.0.1');
    }
```

The expected exception will not be thrown. Instead of it will throw a TypeError exception.

After my modification the test will pass.

boesing: Can you please commit this failing test to the repository? Please avoid using the `StorageFactory` (which will be removed with v3) and instead simply use:

```php
$storage = new Filesystem();
$storage->addPlugin(new Serializer());
```

mariojrrc: @boesing I made the requested changes.

src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -11,6 +11,7 @@
 use Laminas\Cache\Storage\ClearByNamespaceInterface;
 use Laminas\Cache\Storage\FlushableInterface;
 use Laminas\Cache\Storage\StorageInterface;
+use Psr\SimpleCache\CacheException as PsrCacheException;
```
boesing: Could you use `as SimpleCacheExceptionInterface` instead? 
We do use `as SimpleCacheInterface` below so we keep at least some conformity?

Thanks for working on this 👍🏼

boesing (changes requested):
LGTM - just a little nitpick regarding the import statement.

src/Psr/SimpleCache/SimpleCacheDecorator.php
mariojrrc: Sure, thats makes total sense :)

mariojrrc: @boesing is there any other point to make changes?

boesing:
Thanks, @mariojrrc 
Sorry that it took that long.

PR #179, opened 2021-11-16
Documentation: Fix adapter name configuration key

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

In the quick start https://docs.laminas.dev/laminas-cache/v3/storage/adapter/#quick-start section of the docs, it suggests the normalized configuration array requires the adapter name passed as the value to the `adapter` key. Looking at the source https://github.com/laminas/laminas-cache/blob/2bcf8645a9e7a12f6e0332bd9d8be6c228616e55/src/Service/StorageAdapterFactory.php#L39 suggests this should be `name` instead, which is what works currently when testing this.

This PR updates the documentation to reflect the correct key name.

### DIFF
--- a/docs/book/v3/storage/adapter.md
+++ b/docs/book/v3/storage/adapter.md
@@ -52,7 +52,7 @@ $cache = $storageFactory->create(
 
 // Via array configuration:
 $cache = $storageFactory->createFromArrayConfiguration([
-    'adapter' => 'apcu',
+    'name' => 'apcu',
     'options' => ['ttl' => 3600],
     'plugins' => [
         [


froschdesign:
@internalsystemerror 
> …it suggests the normalized configuration array…

This is not a suggestion, just the old documentation for version 2.

Please use [3.0.x branch](https://github.com/laminas/laminas-cache/tree/3.0.x) for this pull request.

internalsystemerror:
@froschdesign done

boesing:
I stumbled over this yesterday as well. That change really wasn't that smart, I think I will allow `adapter` as well since this was the suggestion by the v2.12.0 release... 
I probably failed there... Maybe I have to deprecate `name` afterwards

boesing:
Just FYI: I am closing this, releasing v3.0.1 which adds documentation reference to the migration guide.
Afterwards, I will create a new PR for 3.1.0 which adds `adapter` to the configuration as an optional configuration array key (as thats what v2 supported) and deprecate `name` for the configuration.

This will ensure proper support for v2 migrations as well as usages of https://github.com/laminas/laminas-cache-storage-deprecated-factory

PR #183, opened 2021-11-18
Revert changes made for `adapter` name from `StoragePluginFactoryInterface` and apply to `StorageAdapterFactoryInterface` instead

snapshotpl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Looks like fix #182 introduced by @boesing is wrong. This change should be applied to Adapters, not plugins. That's why https://github.com/doctrine/DoctrineModule/pull/753#issuecomment-971768119 still fails.

### DIFF
--- a/src/Service/StorageAdapterFactory.php
+++ b/src/Service/StorageAdapterFactory.php
@@ -36,7 +36,8 @@ final class StorageAdapterFactory implements StorageAdapterFactoryInterface
 
     public function createFromArrayConfiguration(array $configuration): StorageInterface
     {
-        $adapterName    = $configuration['name'];
+        $adapterName = $configuration['adapter'] ?? $configuration['name'] ?? null;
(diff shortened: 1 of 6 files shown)


froschdesign:
@snapshotpl 
> Looks like fix #182 introduced by @boesing is wrong.

Then the other change should be reversed, right?

snapshotpl:
Yes

snapshotpl:
However then we will introduce new compatibility break...

boesing:
Please target 3.1.1, revert my commit from yesterday and keep these changes.

> However then we will introduce new compatibility break...

These changes were live for less than 24h. Even packagist states 0 installs for 3.1.0 (but its averaged on a weekly basis) so I think we might be fine.

boesing (approved):

PR #185, opened 2021-11-18
Fix storage adapter factory assertion with adapter key

snapshotpl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Fixes missing change in validate method for configuration introduced in #183

### DIFF
--- a/test/Service/StorageAdapterFactoryTest.php
+++ b/test/Service/StorageAdapterFactoryTest.php
@@ -87,21 +87,21 @@ final class StorageAdapterFactoryTest extends TestCase
 
         yield 'missing name' => [
             ['options' => []],
-            'Configuration must contain a "name" key',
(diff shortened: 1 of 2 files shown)


test/Service/StorageAdapterFactoryTest.php
```diff
@@ -119,7 +119,7 @@ protected function setUp(): void
      * @param array<string,mixed> $adapterConfiguration
      * @dataProvider storageConfigurations
      */
-    public function testWillCreateStorageFromArrayConfiguration(
+    public function testWillCreateStorageFromArrayConfigurationWithDeprecatedNameKey(
```
snapshotpl, line 122: Deprecate key test case

PR #199, opened 2022-07-10
`CacheItem` for `PSR-6` needs to start expiring after calling `CacheItemInterface#expiresAfter`

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | kinda
| BC Break      | maybe
| New Feature   | no

### Description

This might introduce some kind of BC break. 
Prior this change, cache items did not start to expire after calling `CacheItemInterface#expiresAfter`.
Therefore, `CacheItemInterface` instances only started to expire after calling one of the following methods:

- `CacheItemPoolInterface#save`
- `CacheItemPoolInterface#commit`

With one of the integration tests we are using for each of our cache adapters, we found out that this is not how cache items should work.

Since `symfony/cache` and `doctrine/cache` do use these integration tests as well, I'd expect that this is a common use-case and thus want to align with these assumptions.

---

Do we want to introduce this slight BC break? I have a few use-cases in mind where one might want to synchronize expiration times of multiple cache items but I would assume that this is not fully breaking a project since whenever one is working with `time`, `microseconds` might be different and thus `CacheItemPoolInterface#getItem` might return items which are not `hit` even tho they should be there from a TTL PoV.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "process-timeout": 600,
         "sort-packages": true,
         "platform": {
-            "php": "7.3.99"
+            "php": "7.4.99"
         },
         "allow-plugins": {
             "dealerdirect/phpcodesniffer-composer-installer": true
@@ -34,7 +34,7 @@
         }
     },
     "require": {
-        "php": "^7.3 || ~8.0.0 || ~8.1.0",
+        "php": "^7.4 || ~8.0.0 || ~8.1.0",
         "laminas/laminas-cache-storage-implementation": "1.0",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-servicemanager": "^3.7",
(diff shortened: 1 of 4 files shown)


boesing:
_Since this change introduces some kind of BC break, I targeted 3.2.x (minor) rather than 3.1.x (patch)._

composer.json
```diff
@@ -44,11 +44,12 @@
         "webmozart/assert": "^1.9"
     },
     "require-dev": {
+        "cache/integration-tests": "dev-master",
```
Ocramius: Should we pin to a specific commit? 🤔

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -164,10 +162,12 @@ public function expiresAfter($time)
...
     {
-        $ttl = $this->ttl;
-        if ($this->expiration !== null) {
-            $ttl = $this->expiration - time();
+        if ($this->expiration === null) {
+            return null;
         }
-        return $ttl;
+
+        $now = new DateTimeImmutable("now", $this->utc);
```
Ocramius, line 169: Not important, but we should consider using a clock in future 🤔

Ocramius (approved):

composer.json
boesing: Is renovate able to update commits? Then yes, why not. 👌🏻

src/Psr/CacheItemPool/CacheItem.php
boesing: Was thinking the same when working on it. 
Will create an issue to keep track of it.

composer.json
Ocramius: I think renovate uses composer 's CLI here, not that advanced 😅

boesing: I removed the requirement from https://github.com/laminas/laminas-cache-storage-adapter-test/releases/tag/2.3.0 so I can drop this requirement here.

Ocramius (approved):

Ocramius:
🚢 the BC concern is very minimal here 👍

PR #207, opened 2022-08-01
Use `Clock` interface for PSR-6 operations

### LINKED ISSUE #200: Use `Clock`-Interface (PSR-20) in `PSR-6` `CacheItem`
Since we do work with times within the `PSR-6` `CacheItem`, we could consider adding the possibility to inject a `Clock`-Interface to the `CacheItemPoolDecorator`.

----

Not important, but we should consider using a clock in future 🤔

_Originally posted by @Ocramius in https://github.com/laminas/laminas-cache/pull/199#discussion_r917465375_

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

This introduces a suggestion from @Ocramius (as tracked in #200) to use a `Clock` for time-based operations.

This also drops support for PHP 7.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,9 +39,9 @@
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-servicemanager": "^3.11",
         "laminas/laminas-stdlib": "^3.6",
-        "lcobucci/clock": "^2.0",
         "psr/cache": "^1.0",
         "psr/simple-cache": "^1.0",
+        "stella-maris/clock": "^0.1.5",
         "webmozart/assert": "^1.9"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -34,11 +34,12 @@
         }
     },
     "require": {
-        "php": "^7.4 || ~8.0.0 || ~8.1.0",
+        "php": "~7.4.0 || ~8.0.0 || ~8.1.0",
         "laminas/laminas-cache-storage-implementation": "1.0",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-servicemanager": "^3.11",
         "laminas/laminas-stdlib": "^3.6",
+        "lcobucci/clock": "^2.0",
         "psr/cache": "^1.0",
         "psr/simple-cache": "^1.0",
         "webmozart/assert": "^1.9"
--- a/docs/book/v3/psr6.md
+++ b/docs/book/v3/psr6.md
@@ -141,9 +141,9 @@ do not need to use a `Laminas\Cache\Storage\Plugin\Serializer` plugin.
 
 ## Time-Related Operations
 
-By default, the PSR-6 cache decorator uses the systems timezone to determine cache expirations. This can be changed by 
+By default, the PSR-6 cache decorator uses the systems timezone to determine cache expirations. This can be changed by
 passing an `lcobucci/clock` `Clock` implementation to the `CacheItemPoolDecorator#__construct`.
-By doing so, the method `CacheItemInterface#expiresAfter` uses that `Clock` to calculate the cache items TTL when passing 
+By doing so, the method `CacheItemInterface#expiresAfter` uses that `Clock` to calculate the cache items TTL when passing
 a `DateInterval` instance.
 
 ```php
--- a/docs/book/v3/psr6.md
+++ b/docs/book/v3/psr6.md
@@ -141,8 +141,8 @@ do not need to use a `Laminas\Cache\Storage\Plugin\Serializer` plugin.
 
 ## Time-Related Operations
 
-By default, the PSR-6 cache decorator uses the systems timezone to determine cache expirations. This can be changed by
-passing an `lcobucci/clock` `Clock` implementation to the `CacheItemPoolDecorator#__construct`.
+By default, the PSR-6 cache decorator uses the systems time to determine cache expirations. This can be changed by
+passing a [`stella-maris/clock`](https://packagist.org/packages/stella-maris/clock) `Clock` implementation to the `CacheItemPoolDecorator#__construct`.
 By doing so, the method `CacheItemInterface#expiresAfter` uses that `Clock` to calculate the cache items TTL when passing
 a `DateInterval` instance.
 
--- a/docs/book/v3/psr6.md
+++ b/docs/book/v3/psr6.md
@@ -138,3 +138,24 @@ returned as a value with exactly the same type.
 Not all adapters can natively store all these types. For instance, Redis stores booleans and integers as a string. Where
 this is the case *all* values will be automatically run through `serialize()` on save and `unserialize()` on get: you
 do not need to use a `Laminas\Cache\Storage\Plugin\Serializer` plugin.
+
+## Time-Related Operations
+
+By default, the PSR-6 cache decorator uses the systems timezone to determine cache expirations. This can be changed by 
+passing an `lcobucci/clock` `Clock` implementation to the `CacheItemPoolDecorator#__construct`.
+By doing so, the method `CacheItemInterface#expiresAfter` uses that `Clock` to calculate the cache items TTL when passing 
+a `DateInterval` instance.
+
+```php
+use Lcobucci\Clock\SystemClock;
+$clock = new SystemClock('Antarctica/Troll');
+$pool = new CacheItemPoolDecorator($storage, $clock);
+
+$item = $pool->getItem('non-existent-key');
+$item
+    ->set('foo')
+    ->expiresAfter(DateInterval::createFromDateString('24 hours'));
+
+// Saves the item which expires in 24h
+$pool->save($item);
+```
--- a/src/Psr/CacheItemPool/CacheItem.php
+++ b/src/Psr/CacheItemPool/CacheItem.php
@@ -3,10 +3,10 @@
 namespace Laminas\Cache\Psr\CacheItemPool;
 
 use DateInterval;
+use DateTimeImmutable;
 use DateTimeInterface;
-use Lcobucci\Clock\Clock;
-use Lcobucci\Clock\SystemClock;
 use Psr\Cache\CacheItemInterface;
+use StellaMaris\Clock\ClockInterface;
 
 use function gettype;
 use function is_int;
@@ -42,17 +42,23 @@ final class CacheItem implements CacheItemInterface
      */
     private ?int $expiration = null;
 
-    private Clock $clock;
+    private ClockInterface $clock;
 
     /**
      * @param mixed $value
      */
-    public function __construct(string $key, $value, bool $isHit, ?Clock $clock = null)
+    public function __construct(string $key, $value, bool $isHit, ?ClockInterface $clock = null)
     {
         $this->key   = $key;
         $this->value = $isHit ? $value : null;
         $this->isHit = $isHit;
-        $clock     ??= SystemClock::fromSystemTimezone();
+        $clock     ??= new class implements ClockInterface
+        {
+            public function now(): DateTimeImmutable
+            {
+                return new DateTimeImmutable();
+            }
+        };
         $this->clock = $clock;
     }
 
--- a/src/Psr/CacheItemPool/CacheItem.php
+++ b/src/Psr/CacheItemPool/CacheItem.php
@@ -3,9 +3,9 @@
 namespace Laminas\Cache\Psr\CacheItemPool;
 
 use DateInterval;
-use DateTimeImmutable;
 use DateTimeInterface;
-use DateTimeZone;
+use Lcobucci\Clock\Clock;
+use Lcobucci\Clock\SystemClock;
 use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
 
@@ -43,19 +43,20 @@ final class CacheItem implements CacheItemInterface
      */
     private ?int $expiration = null;
 
-    private DateTimeZone $utc;
+    private Clock $clock;
 
     /**
      * @param string $key
      * @param mixed $value
      * @param bool $isHit
      */
-    public function __construct($key, $value, $isHit)
+    public function __construct($key, $value, $isHit, ?Clock $clock = null)
     {
         $this->key   = $key;
         $this->value = $isHit ? $value : null;
         $this->isHit = $isHit;
-        $this->utc   = new DateTimeZone('UTC');
+        $clock     ??= SystemClock::fromSystemTimezone();
+        $this->clock = $clock;
     }
 
     /**
@@ -146,7 +147,7 @@ final class CacheItem implements CacheItemInterface
 
         /** @psalm-suppress RedundantConditionGivenDocblockType Until we do have native type-hints we should keep verifying this. */
         if ($time instanceof DateInterval) {
-            $now = new DateTimeImmutable('now', $this->utc);
+            $now = $this->clock->now();
             return $this->expiresAt($now->add($time));
         }
 
@@ -166,7 +167,7 @@ final class CacheItem implements CacheItemInterface
             return null;
         }
 
-        $now = new DateTimeImmutable("now", $this->utc);
+        $now = $this->clock->now();
 
         return $this->expiration - $now->getTimestamp();
     }
--- a/src/Psr/CacheItemPool/CacheItem.php
+++ b/src/Psr/CacheItemPool/CacheItem.php
@@ -7,11 +7,18 @@ use DateTimeImmutable;
 use DateTimeInterface;
 use DateTimeZone;
 use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
 
 use function gettype;
 use function is_int;
 use function sprintf;
 
+/**
+ * @internal The cache item should only be used by this component. To create one or more new cache item, use
+ *           {@see \Psr\Cache\CacheItemPoolInterface::getItem()} or {@see \Psr\Cache\CacheItemPoolInterface::getItems()}
+ *           instead. These methods will provide one or more {@see CacheItemInterface} instances which can be modified
+ *           by using the methods declared in the interface.
+ */
 final class CacheItem implements CacheItemInterface
 {
     /**
--- a/src/Psr/CacheItemPool/CacheItem.php
+++ b/src/Psr/CacheItemPool/CacheItem.php
@@ -7,7 +7,6 @@ use DateTimeInterface;
 use Lcobucci\Clock\Clock;
 use Lcobucci\Clock\SystemClock;
 use Psr\Cache\CacheItemInterface;
-use Psr\Cache\CacheItemPoolInterface;
 
 use function gettype;
 use function is_int;
@@ -36,7 +35,7 @@ final class CacheItem implements CacheItemInterface
     /**
      * True if the cache item lookup resulted in a cache hit or if they item is deferred or successfully saved
      */
-    private bool $isHit = false;
+    private bool $isHit;
 
     /**
      * Timestamp item will expire at if expiresAt() called, null otherwise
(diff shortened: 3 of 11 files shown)


src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -3,24 +3,27 @@
 namespace Laminas\Cache\Psr\CacheItemPool;
 
 use DateInterval;
-use DateTimeImmutable;
 use DateTimeInterface;
-use DateTimeZone;
+use Lcobucci\Clock\Clock;
```
heiglandreas: Could we perhaps use `stella-maris/clock` as interface? That will allow a bit more interoperability as users can then decide which of the implementations they might want to use. As that interface is forward-compatible to the current draft of `psr/clock` that should also make this a future-proof solution. 

As `lcobbuci/clock' implements that interface there is not much change necessary...

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -3,24 +3,27 @@
...
+use Lcobucci\Clock\SystemClock;
```
heiglandreas: As fallback solution I'd rather use a separate implementation that we ship with this package instead of a third-party as that would require that third party to actually be used.

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -8,6 +8,8 @@
 use Laminas\Cache\Storage\ClearByNamespaceInterface;
 use Laminas\Cache\Storage\FlushableInterface;
 use Laminas\Cache\Storage\StorageInterface;
+use Lcobucci\Clock\Clock;
+use Lcobucci\Clock\SystemClock;
```
heiglandreas: See above notes.

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -1052,4 +1053,22 @@ public function testKeepsDeferredItemsWhenCommitFails(): void
         self::assertTrue($this->adapter->hasItem('keyOfFailedItem1'));
         self::assertTrue($this->adapter->hasItem('keyOfFailedItem2'));
     }
+
+    public function testPassesClockToCacheItem(): void
+    {
+        $clock   = $this->createMock(Clock::class);
```
heiglandreas: We should not mock a clock but instead use the corresponding `FrozenClock` instead

composer.json
```diff
@@ -34,21 +34,23 @@
...
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-servicemanager": "^3.7",
         "laminas/laminas-stdlib": "^3.6",
+        "lcobucci/clock": "^2.0",
         "psr/cache": "^1.0",
         "psr/simple-cache": "^1.0",
         "webmozart/assert": "^1.9"
     },
     "require-dev": {
+        "cache/integration-tests": "dev-master",
```
Ocramius: Hmm, we use `dev-master` only on this package? :thinking:

test/Psr/CacheItemPool/CacheItemTest.php
```diff
@@ -106,4 +110,32 @@ public function testExpiresAfterInvalidThrowsException(): void
...
+        self::assertEquals(9, $item->getTtl());
+    }
+
+    public function testClockProvidedDoesNotContainUTCTimeZone(): void
+    {
+        $item = new CacheItem(
+            'foo',
+            null,
+            false,
+            new SystemClock(new DateTimeZone('Europe/Berlin'))
```
heiglandreas: In tests it usually makes more sense to use the `FrozenClock` instead of the actual one.

heiglandreas (changes requested):

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -31,45 +34,32 @@ final class CacheItem implements CacheItemInterface
...
      */
-    public function __construct($key, $value, $isHit)
+    public function __construct(string $key, $value, bool $isHit, ?Clock $clock = null)
     {
         $this->key   = $key;
         $this->value = $isHit ? $value : null;
         $this->isHit = $isHit;
-        $this->utc   = new DateTimeZone('UTC');
+        $clock     ??= SystemClock::fromSystemTimezone();
+        $this->clock = $clock;
```
Ocramius: Having the clock on a `CacheItem` seems weird to me: this makes a `CacheItem` a mutable/stateful object that depends on global state :thinking:

Ocramius: `getTtl()` makes it kinda required to have this design.

test/Psr/CacheItemPool/CacheItemTest.php
```diff
@@ -106,4 +110,32 @@ public function testExpiresAfterInvalidThrowsException(): void
         $item = new CacheItem('key', 'value', true);
         $item->expiresAfter([]);
     }
+
+    public function testExpiresAfterStartsExpiringAfterMethodCall(): void
+    {
+        $item = new CacheItem('key', 'value', false);
+        $item = $item->expiresAfter(10);
+        sleep(1);
```
Ocramius: Shouldn't be needed, with the new clock? Or are you explicitly testing the system clock default behavior here?

test/Psr/CacheItemPool/CacheItemTest.php
```diff
@@ -106,4 +110,32 @@ public function testExpiresAfterInvalidThrowsException(): void
...
+
+        $interval = DateInterval::createFromDateString('1 hour');
+        $item->expiresAfter($interval);
+
+        self::assertEquals(3600, $item->getTtl());
+
+        $inOneHour = (new DateTimeImmutable('now', new DateTimeZone('America/Vancouver')))->add($interval);
+        $item->expiresAt($inOneHour);
+
+        self::assertEquals(3600, $item->getTtl());
```
Ocramius, line 151: This test is a bit fragile: may fail if `SystemClock` moves while we're observing this (for whatever reason, the test has a hiccup)

Ocramius (approved):
LGTM - improvements in the test suite can be performed later too.

composer.json
```diff
@@ -34,21 +34,23 @@
         }
     },
     "require": {
-        "php": "^7.3 || ~8.0.0 || ~8.1.0",
+        "php": "~7.4.0 || ~8.0.0 || ~8.1.0",
         "laminas/laminas-cache-storage-implementation": "1.0",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-servicemanager": "^3.7",
         "laminas/laminas-stdlib": "^3.6",
+        "lcobucci/clock": "^2.0",
```
heiglandreas: Let's not depend on one single implementation. `stella-maris/clock` provides an interface implemented by multiple different clock-liberaries. `lcobucci/clock` is one of them.

composer.json
boesing: its being removed with #199 
Since this work was based on early #199 state, I will rebase after #199 is merged

composer.json
heiglandreas:
```suggestion
        "stella-maris/clock": "^1.0",
```

composer.json
heiglandreas:
```suggestion
        "cache/integration-tests": "dev-master",
        "lcobucci/clock": "^2.0",
```

docs/book/v3/psr6.md
```diff
@@ -138,3 +138,24 @@ returned as a value with exactly the same type.
 Not all adapters can natively store all these types. For instance, Redis stores booleans and integers as a string. Where
 this is the case *all* values will be automatically run through `serialize()` on save and `unserialize()` on get: you
 do not need to use a `Laminas\Cache\Storage\Plugin\Serializer` plugin.
+
+## Time-Related Operations
+
+By default, the PSR-6 cache decorator uses the systems timezone to determine cache expirations. This can be changed by
+passing an `lcobucci/clock` `Clock` implementation to the `CacheItemPoolDecorator#__construct`.
+By doing so, the method `CacheItemInterface#expiresAfter` uses that `Clock` to calculate the cache items TTL when passing
+a `DateInterval` instance.
```
heiglandreas: I'm not 100% sure we actually need a clock for this but there might be edge-cases where a TTL of for example `7 days` might get different results depending on DST transitions. Though I'm not really sure that makes that much of a difference for user-land. For the testing purposes it definitely makes sense as we can skip calling `sleep` as we can return different datetimes in different calls to `now` so we can "simulate" passing of a certain amount of time. 

```suggestion
By default, the PSR-6 cache decorator uses the systems time to determine cache expirations. This can be changed by
passing a [`stella-maris/clock`](https://packagist.org/packages/stella-maris/clock) `Clock` implementation to the `CacheItemPoolDecorator#__construct`.
By doing so, the method `CacheItemInterface#expiresAfter` uses that `Clock` to calculate the cache items TTL when passing
a `DateInterval` instance.
```

src/Psr/CacheItemPool/CacheItem.php
heiglandreas:
```suggestion
use StellaMaris\Clock\ClockInterface;
```

src/Psr/CacheItemPool/CacheItem.php
heiglandreas:
```suggestion
```

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -31,45 +34,32 @@ final class CacheItem implements CacheItemInterface
...
-    /**
-     * Seconds after item is stored it will expire at if expiresAfter() called, null otherwise
-     *
-     * @var int|null
-     */
-    private $ttl;
-
-    /** @var DateTimeZone */
-    private $utc;
+    private Clock $clock;
```
heiglandreas:
```suggestion
    private ClockInterface $clock;
```

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -31,45 +34,32 @@ final class CacheItem implements CacheItemInterface
...
-     * @param bool $isHit
      */
-    public function __construct($key, $value, $isHit)
+    public function __construct(string $key, $value, bool $isHit, ?Clock $clock = null)
     {
         $this->key   = $key;
         $this->value = $isHit ? $value : null;
         $this->isHit = $isHit;
-        $this->utc   = new DateTimeZone('UTC');
+        $clock     ??= SystemClock::fromSystemTimezone();
```
heiglandreas:
```suggestion
        $clock     ??= new class implements ClockInterface { function now() { return new DateTimeImmutable(); } };
```

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
heiglandreas:
```suggestion
use StellaMaris\Clock\ClockInterface;
```

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -41,11 +43,12 @@ class CacheItemPoolDecorator implements CacheItemPoolInterface
...
 
-    /** @var StorageInterface */
-    private $storage;
+    private StorageInterface $storage;
 
     /** @var array<string,CacheItem> */
-    private $deferred = [];
+    private array $deferred = [];
+
+    private Clock $clock;
```
heiglandreas:
```suggestion
    private ClockInterface $clock;
```

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -54,12 +57,14 @@ class CacheItemPoolDecorator implements CacheItemPoolInterface
...
      * @throws CacheException
      */
-    public function __construct(StorageInterface $storage)
+    public function __construct(StorageInterface $storage, ?Clock $clock = null)
     {
         $this->validateStorage($storage);
         $capabilities = $storage->getCapabilities();
         $this->memoizeMaximumKeyLengthCapability($storage, $capabilities);
         $this->storage = $storage;
+        $clock       ??= SystemClock::fromSystemTimezone();
```
heiglandreas:
```suggestion
        $clock       ??= new class implements ClockInterface { function now() { return new DateTimeImmutable(); } }
```

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -17,6 +17,7 @@
 use Laminas\EventManager\EventManager;
 use LaminasTest\Cache\Psr\CacheItemPool\TestAsset\FlushableStorageAdapterInterface;
 use LaminasTest\Cache\Psr\TestAsset\FlushableNamespaceStorageInterface;
+use Lcobucci\Clock\Clock;
```
heiglandreas:
```suggestion
use Lcobucci\Clock\FrozenClock;
```

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
heiglandreas:
```suggestion
        $now = new DateTimeImmutable();
        $clock   = new FrozenClock($now);
```

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -1052,4 +1053,22 @@ public function testKeepsDeferredItemsWhenCommitFails(): void
...
+        $item    = $adapter->getItem('notExistingItem');
+        self::assertInstanceOf(CacheItem::class, $item);
+        self::assertFalse($item->isHit());
+
+        $now = new DateTimeImmutable('now');
+        $clock
+            ->expects(self::once())
+            ->method('now')
+            ->willReturn($now);
+
```
heiglandreas, line 1092:
```suggestion
```

test/Psr/CacheItemPool/CacheItemTest.php
```diff
@@ -4,12 +4,16 @@
 
 use DateInterval;
 use DateTime;
+use DateTimeImmutable;
+use DateTimeZone;
 use Laminas\Cache\Psr\CacheItemPool\CacheItem;
 use Laminas\Cache\Psr\CacheItemPool\InvalidArgumentException;
+use Lcobucci\Clock\SystemClock;
```
heiglandreas:
```suggestion
use Lcobucci\Clock\FrozenClock;
```

test/Psr/CacheItemPool/CacheItemTest.php
```diff
@@ -106,4 +110,32 @@ public function testExpiresAfterInvalidThrowsException(): void
...
         $item->expiresAfter([]);
     }
+
+    public function testExpiresAfterStartsExpiringAfterMethodCall(): void
+    {
+        $item = new CacheItem('key', 'value', false);
+        $item = $item->expiresAfter(10);
+        sleep(1);
+        self::assertEquals(9, $item->getTtl());
+    }
```
heiglandreas:
```suggestion
    public function testExpiresAfterStartsExpiringAfterMethodCall(): void
    {
        $clock = new FrozenClock(new DateTimeImmutable('@1'));
        $item = new CacheItem('key', 'value', false, $clock);
        $item = $item->expiresAfter(10);
        $clock->set(new DateTimeImmutable('@2'));
        self::assertEquals(9, $item->getTtl());
    }
```

test/Psr/CacheItemPool/CacheItemTest.php
```diff
@@ -106,4 +110,32 @@ public function testExpiresAfterInvalidThrowsException(): void
...
+        $interval = DateInterval::createFromDateString('1 hour');
+        $item->expiresAfter($interval);
+
+        self::assertEquals(3600, $item->getTtl());
+
+        $inOneHour = (new DateTimeImmutable('now', new DateTimeZone('America/Vancouver')))->add($interval);
+        $item->expiresAt($inOneHour);
+
+        self::assertEquals(3600, $item->getTtl());
+    }
```
heiglandreas:
```suggestion
    public function testClockProvidedDoesNotContainUTCTimeZone(): void
    {
        $berlin = new DateTimeImmutable('2022-10-30 02:30:00', new DateTimezone('Europe/Berlin'));
        $clock = new FrozenClock($berlin);
        
        $item = new CacheItem('foo', null, false, $clock);

        $interval = DateInterval::createFromDateString('1 hour');
        $item->expiresAfter($interval);

        self::assertEquals(3600, $item->getTtl());

        $inOneHour = (new DateTimeImmutable('2022-10-29 18:30:00', new DateTimeZone('America/Vancouver')))->add($interval);
        $item->expiresAt($inOneHour);

        self::assertEquals(3600, $item->getTtl());
    }
```

heiglandreas (commented):
Think I got all parts covered.

composer.json
Ocramius: Discussed on slack: using `stella-maris/clock` is a sufficient solution for now.

1. `lcobucci/clock` implements `stella-maris/clock` for now
2. we can provide our own clock with a minimal anonymous class: `class implements Clock { function now(): DateTimeImmutable { return new DateTimeImmutable(); }}`

This reduces the number of installed dependencies by one.

Also, I didn't know that `lcobucci/clock` required (and implemented) `stella-maris/clock`, at the time of my first review.

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
boesing: I do explicitly want to ensure that the method is called. Thats why I prefer the mock here. We can use frozen clock in other scenarios for sure but since I removed `lcobucci/clock` in favor of `stella-maris` which does not provide such a frozen clock, I guess I keep working with the mock for now.

test/Psr/CacheItemPool/CacheItemTest.php
boesing: I am open for another solution, I spent quite a while today and don't come to a good conclusion tho. Maybe tomorrow.

boesing:
@Ocramius @heiglandreas I've applied everything related to `stella-maris/clock`.
I also removed the `sleep` from the `CacheItemTest` which is not necessary anymore after implementing the `Clock`. The test was provided by another PR and I just haven't updated it. Good finding tho.

Regarding the timezone check, I have no idea how to optimize that check tho. 
Is it suitable to use a very specific time rather than using `now`? Sorry, I spent 4 hours in train after an 8h workday and thus my brain stopped working properly 😂

Ocramius:
I think we can leave the timezone stuff out at first, and optimize in a new minor, since we don't expose the used `DateTimeImmutable` instances to the outside.

docs/book/v3/psr6.md
```diff
@@ -138,3 +138,24 @@ returned as a value with exactly the same type.
...
+## Time-Related Operations
+
+By default, the PSR-6 cache decorator uses the systems time to determine cache expirations. This can be changed by
+passing a [`stella-maris/clock`](https://packagist.org/packages/stella-maris/clock) `Clock` implementation to the `CacheItemPoolDecorator#__construct`.
+By doing so, the method `CacheItemInterface#expiresAfter` uses that `Clock` to calculate the cache items TTL when passing
+a `DateInterval` instance.
+
+```php
+use Lcobucci\Clock\SystemClock;
+$clock = new SystemClock('Antarctica/Troll');
```
Ocramius, line 151: Best timezone

test/Psr/CacheItemPool/CacheItemTest.php
Ocramius: You fixed this 👍

Ocramius (approved):
🚢  nice work, @boesing!

heiglandreas:
To me that's fine as the clocks `now` will make sure that the production code uses the right time

PR #209, opened 2022-08-11
Apply PHP 7.4 syntax and typed property

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since composer.json require php 7.4, php 7.4 syntax can be applied with typed properties. 

- For `final class`, typed properties for protected is allowed if no inherit of parent or used by its final class parent.
- Otherwise, update only private properties.

### DIFF
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -51,8 +51,6 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      * operation.
      *
      * @var null|bool
-     *
-     * Use ?bool type make it unserialized, even with null initialized
      */
     private $success;
 
--- a/src/Psr/SimpleCache/SimpleCacheDecorator.php
+++ b/src/Psr/SimpleCache/SimpleCacheDecorator.php
@@ -42,22 +42,21 @@ class SimpleCacheDecorator implements SimpleCacheInterface
      */
     public const INVALID_KEY_CHARS = ':@{}()/\\';
 
-    /** @var bool */
-    private $providesPerItemTtl = true;
+    private bool $providesPerItemTtl = true;
 
-    /** @var StorageInterface */
-    private $storage;
+    private StorageInterface $storage;
 
     /**
      * Reference used by storage when calling getItem() to indicate status of
      * operation.
      *
      * @var null|bool
+     *
+     * Use ?bool type make it unserialized, even with null initialized
      */
     private $success;
 
-    /** @var DateTimeZone */
-    private $utc;
+    private DateTimeZone $utc;
 
     public function __construct(StorageInterface $storage)
     {
@@ -92,7 +91,7 @@ class SimpleCacheDecorator implements SimpleCacheInterface
             throw static::translateThrowable($e);
         }
 
-        $result = $result ?? $default;
+        $result ??= $default;
         return $this->success ? $result : $default;
     }
 
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -928,7 +928,7 @@
       <code>$event</code>
     </MissingClosureParamType>
     <MissingClosureReturnType occurrences="1">
-      <code>function ($event) use ($retVal) {</code>
+      <code>static function ($event) use ($retVal) {</code>
     </MissingClosureReturnType>
     <MixedArgument occurrences="1">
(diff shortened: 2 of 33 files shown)


samsonasik:
All green 🎉

src/Psr/SimpleCache/SimpleCacheDecorator.php
```diff
@@ -42,22 +42,21 @@ class SimpleCacheDecorator implements SimpleCacheInterface
...
-    private $storage;
+    private StorageInterface $storage;
 
     /**
      * Reference used by storage when calling getItem() to indicate status of
      * operation.
      *
      * @var null|bool
+     *
+     * Use ?bool type make it unserialized, even with null initialized
```
Ocramius: This is confusing 🤔

Ocramius (changes requested):

src/Psr/SimpleCache/SimpleCacheDecorator.php
samsonasik: Yes, I don't know what possible values are /cc @boesing

samsonasik: with `?bool $success = null`, it got error:

```bash
There were 11 errors:

1) LaminasTest\Cache\Psr\SimpleCache\SimpleCacheIntegrationTest::testSetTtl
Laminas\Cache\Psr\SimpleCache\SimpleCacheException: Serialized data must be a string containing serialized PHP code; received: boolean

/Users/samsonasik/www/laminas-cache/src/Psr/SimpleCache/SimpleCacheDecorator.php:322
/Users/samsonasik/www/laminas-cache/src/Psr/SimpleCache/SimpleCacheDecorator.php:91
/Users/samsonasik/www/laminas-cache/vendor/laminas/laminas-cache-storage-adapter-test/src/AbstractSimpleCacheIntegrationTest.php:220
```

boesing: I am not sure if I understand what you try to outline with `unserialized`.
What I can say is that `laminas-cache` has mostly no typing and is passing plenty of stuff by-reference.
I do not really "trust" upstream as of now and thus, can't say if `success` is always boolean.

Therefore, I'd say lets keep it as is.
But I definitely do not get what the point of the comment is.

samsonasik: @boesing ok, removed comment https://github.com/laminas/laminas-cache/pull/209/commits/79a244dda045207056b366df65fea41bc707f2d6 👍

boesing: Can you please tell me what you wanted to point out with the comment? 
Because I just did not understood it.
Its not that I wanted to get rid of it, just wanted to have some more backstory 😅 

E: Oh, I see. I started writing my comment before you sent the output of the error...

samsonasik: It has `@var null|bool` in the docblock, so it assume it should be `?bool`, and if it doesn't has initialization in `__construct`, it means it needs to have default `null` value, so the type should be:

```php
private ?bool $success = null;
```

boesing: > with `?bool $success = null`, it got error:
> 
> ```shell
> There were 11 errors:
> 
> 1) LaminasTest\Cache\Psr\SimpleCache\SimpleCacheIntegrationTest::testSetTtl
> Laminas\Cache\Psr\SimpleCache\SimpleCacheException: Serialized data must be a string containing serialized PHP code; received: boolean
> 
> /Users/samsonasik/www/laminas-cache/src/Psr/SimpleCache/SimpleCacheDecorator.php:322
> /Users/samsonasik/www/laminas-cache/src/Psr/SimpleCache/SimpleCacheDecorator.php:91
> /Users/samsonasik/www/laminas-cache/vendor/laminas/laminas-cache-storage-adapter-test/src/AbstractSimpleCacheIntegrationTest.php:220
> ```


TBH: I have no idea what this error is about. Maybe its worth digging into it? But anyways, keeping the code as is might be suitable for now. I'd rather go without this one property optimized for PHP 7.4 than breaking upstream code because of this. 👍🏼

boesing (approved):
LGTM!

PR #216, opened 2022-08-31
Add support for psr/cache 2 and 3 versions

rimvislt:


Ocramius:
FWIW, let's drop PHP 7.4 as part of this? It has a shelf life of ~2 months left.

rimvislt:
I was thinking if it is easier to drop php7 or try to load different files based on php version.

rimvislt:
All test are passing minimum version is php8.0 now, let me know if anything else needed before this can be merged and released.

rimvislt:
I am not sure about DCO not passing if I should do it or administrators. Let me know if I need to do anything otherwise everything else are passing now.

.laminas-ci.json
```diff
@@ -1,10 +1,45 @@
 {
     "ignore_php_platform_requirements": {
-        "8.1": true
+        "8.0": false,
```
boesing: We can omit these here as with one of the latest versions of the matrix action I've changed the defaults of all versions to `false`

.laminas-ci.json
```diff
@@ -1,10 +1,45 @@
...
     },
     "extensions": [
         "apcu"
     ],
+    "additional_checks": [
+        {
+            "name": "Unit tests with psr/cache v2",
+            "job": {
+                "php": "8.0",
+                "dependencies": "latest",
```
boesing: should be possible to have `*` here so we can reduce the amount of entries to 2 here. 👍🏼

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -106,7 +106,7 @@ public function setIsHit(bool $isHit): self
     /**
      * {@inheritdoc}
      */
-    public function set($value): CacheItemInterface
+    public function set($value): static
```
boesing: Can we also add argument types so we are aligned with the interface?

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -116,7 +116,7 @@ public function set($value): CacheItemInterface
     /**
      * {@inheritdoc}
      */
-    public function expiresAt($expiration): CacheItemInterface
+    public function expiresAt($expiration): static
```
boesing, line 119: Can we also add argument types so we are aligned with the interface?

src/Psr/CacheItemPool/CacheItem.php
```diff
@@ -130,7 +130,7 @@ public function expiresAt($expiration): CacheItemInterface
     /**
      * {@inheritdoc}
      */
-    public function expiresAfter($time): CacheItemInterface
+    public function expiresAfter($time): static
```
boesing, line 133: Can we also add argument types so we are aligned with the interface?

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -84,7 +83,7 @@ public function __destruct()
     /**
      * {@inheritdoc}
      */
-    public function getItem($key)
+    public function getItem($key): CacheItemInterface
```
boesing, line 86: Can we also add argument types so we are aligned with the interface?

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -149,7 +148,7 @@ public function getItems(array $keys = [])
     /**
      * {@inheritdoc}
      */
-    public function hasItem($key)
+    public function hasItem($key): bool
```
boesing, line 151: Can we also add argument types so we are aligned with the interface?

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -196,15 +195,15 @@ public function clear()
     /**
      * {@inheritdoc}
      */
-    public function deleteItem($key)
+    public function deleteItem($key): bool
```
boesing, line 198: Can we also add argument types so we are aligned with the interface?

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
```diff
@@ -295,6 +301,9 @@ public function testGetItemsRuntimeExceptionIsMiss(): void
             ->with($keys)
             ->willThrowException(new Exception\RuntimeException());
 
+        /**
+         * @var array<string, CacheItem> $items
```
boesing, line 306: do we have to provide these here?

boesing (changes requested):
Nice! Thanks for working on this.
Just a few things but overall, good work 👍🏼

.laminas-ci.json
boesing: Ah, we should also add `*` for `php` since we now dropped PHP 7.4 support and thus can test all versions

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
rimvislt: Yes because psalm is complaining that can not understand data, since response is "iterable"

rimvislt: Yes, psalm does not work because original documentation inside interface gives just return type iterable and it is not enough. I tried to add it on implemented methods but psalm does not work with 2 return types defined (did not want to loose original doc type)

.laminas-ci.json
rimvislt: fixed

.laminas-ci.json
rimvislt: fixed

src/Psr/CacheItemPool/CacheItem.php
rimvislt: added (needed to drop "psr/cache": "^1.0  for it to work)

src/Psr/CacheItemPool/CacheItem.php
rimvislt: added (needed to drop "psr/cache": "^1.0  for it to work)

src/Psr/CacheItemPool/CacheItem.php
rimvislt: added (needed to drop "psr/cache": "^1.0  for it to work)

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
rimvislt: added (needed to drop "psr/cache": "^1.0  for it to work)

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
rimvislt: added (needed to drop "psr/cache": "^1.0  for it to work)

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
rimvislt: added (needed to drop "psr/cache": "^1.0  for it to work)

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
```diff
@@ -84,7 +83,7 @@ public function __destruct()
     /**
      * {@inheritdoc}
      */
-    public function getItem($key)
+    public function getItem(string $key): CacheItemInterface
```
Ocramius: Because this class is not `final`, the change in interface also implicitly becomes a BC break for people not depending on `psr/cache`.

@boesing that means new major version here too :S

phpcs.xml
```diff
@@ -17,5 +17,8 @@
     <file>benchmark</file>
 
     <!-- Include all rules from the Laminas Coding Standard -->
-    <rule ref="LaminasCodingStandard"/>
+    <rule ref="LaminasCodingStandard">
+        <!-- Exclude declare(strict_types=1) until the next major release. This component heavily depends on type-magic and thus should avoid adding strict_types as of now v3 -->
```
Ocramius, line 21: If we roll a new major release, it would be fine to remove this exclusion.

As the current patch is presented, a new major version is indeed needed.

src/Psr/CacheItemPool/CacheItem.php
Ocramius: @boesing if you want the parameter types to match the interface, we will need a new major version here.

If we can keep `psr/cache:^1` and BC, then we can release a minor.

phpcs.xml
rimvislt: If we do major release we will need to add it into "doctrine/doctrine-module" it depends on v3 ->   "laminas/laminas-cache": "^3.1.2".

Let me know what is final decision on this and I will update it accordingly.

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
froschdesign: ~I think a major version is not needed at this point. The changes to the class `CacheItemPoolDecorator` can be undone, then a minor version will also work.~

Version 2 and 3 of the PSR-6 interfaces add type hints. (unfortunately they are not listed in the documentation)
Therefore, my comment here does not fit.

froschdesign: I think a major version is not needed at this point. The changes to the class `CacheItemPoolDecorator` can be undone, then a minor version will also work.

froschdesign: I think a major version is not needed at this point. The changes to the class `CacheItemPoolDecorator` can be undone, then a minor version will also work.

froschdesign: I think a major version is not needed at this point. The changes to the class `CacheItemPoolDecorator` can be undone, then a minor version will also work.

Ocramius: Tend to agree: new major can happen later, if we want :+1:

rimvislt: Removed typed arguments and added back psr/cache v1 support.

phpcs.xml
Ocramius: Decided not to roll with a new major: resolving here.

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
Ocramius: Adding the return type here is still a BC break :| 

We're still in a "new major target" scenario, I fear...

rimvislt: Most likely previously it was just return type in documentation but support for psr/cache v2 and v3 it needs to be added.

Ocramius: Yup: not blaming your changes, just saying that the lack of `final` has exposed us to BC issues here, since the dependency change leads to a transitive BC break.

froschdesign: I see no changes to support all three versions of PSR-6. Therefore it would be good to have plan before things are constantly being changed here.
@boesing, you as maintainer of this component, what is your plan for a new major version?

boesing: Oh, I've expected the decorators to be final since v3.
Seems that I forgot these...

I would be open for having a new major but only if we also add support for `psr/simple-cache` in its newer versions.

In the next major we can add `final` to the decorators as well. 

So we do have to drop v1 support of the psr packages as well in the next major? bummer...

test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php
boesing: Oh I see. Thats because they do not provide what is inside the iterable.
I wonder if its possible to add such stuff to the PSR packages...

src/Psr/CacheItemPool/CacheItemPoolDecorator.php
froschdesign: > I would be open for having a new major but only if we also add support for `psr/simple-cache` in its newer versions.

👍🏻

rimvislt: So what is the plan with this one? Does it mean that v4 will happen in September?

boesing: I do not give an ETA but I will take care that this lands in v4.

Since that requires all supported adapters to support v4 as well plus the fact that there are some other things I've planned for v4, I dont think this will happen in september.

rimvislt: Thank you for your response let me know how I can help with v4.

If you will create v4 branch  I will create new pr against it and will fix this code for it.

rogervila:
Hi @rimvislt could you please fix the branch conflicts? Let me know if I can help on fixing this branch!

rimvislt:
> 

I think above was decided that we will not use this pr and will be new version of it, so not sure if it is a point to fix it.

boesing:
Yah, this PR provides a lot of changes which are not only related to `psr/cache` v3.

We will support `psr/cache` v2 & v3 and `psr/simple-cache` v2 & v3 with one of the next major versions as there is no way of introducing these versions without breaking BC.

----

I'd love to get some prepared PRs so I can simply merge them once we are creating the next major version(s). This will most likely happen after we created `laminas-servicemanager` v4 - even tho there is no official ETA for `servicemanager`, I am planning to create a release in the first quarter of 2023.

The changes in the PRs (please create dedicated PRs for both PSR packages) should contain:

- [ ] interface adaptions (i.e. added argument types along with return types - including `mixed`) within the delegator
- [ ] `composer.json` + `composer.lock` changes regarding the psr-package, this includes both `require` and `provides` keys

Thanks!

PR #220, opened 2022-09-10
Update laminas/laminas-coding-standard to ~2.4.0

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This PR:
- Updates dependencies to their minimum installable versions
- Updates `laminas/laminas-coding-standard` to `~2.4.0` (closes https://github.com/laminas/laminas-cache/pull/203)


Ocramius (changes requested):
I'm a bit on the fence on adding `declare(strict_types=1)` in `src/` here: loads of string operations happening, especially in options :thinking:

internalsystemerror:
There are type hints in the docblocks already, and casting where relevant. I would think psalm would have complained about something if there were any specific issue. Is there something I can add (a unit test maybe) to give us more confidence in this?

Ocramius:
Well, we still have a bit of a baseline there: if the baseline was empty, I'd be happier :P 

Anyway, @boesing's the main maintainer here, so he calls the shots :)

boesing:
Yah, so not gonna happen that we declare strict types here. I will see if laminas-cache v4 actually receives proper types while also dropping PHP <8.0

Adding strict types here will definitely break upstream as there is so much whack stuff going on with type referencing, etc.

Happy to see all the other changes tho, just don't apply the strict types stuff. We can disable that until v4 tho.

internalsystemerror:
I'll can remove the strict type declarations and ignore them altering `phpcs.xml`. My first thought however is, shouldn't there be failing tests if that's the case? I can look at adding those whilst I'm at it.

internalsystemerror:
It took a few attempts, but `declare(strict_types=1)` should be reverted to pre-PR status now.

Ocramius (approved):
Yep, looking good, thanks!

PR #230, opened 2022-10-16
Drop support php 7, add php 8.2 support

fezfez:


fezfez:
missing php8.2-apcu

boesing:
APCu was already installed from source in PHP 8.1 as well, maybe we can just change the check in `pre-install.sh`?
It should be available for PHP 8.1 as of now, so maybe just change that file from `8.1` to `8.2` and it might work.

There is at least a 5.1.22 version which has support for PHP 8.2:
https://github.com/krakjoe/apcu/releases/tag/v5.1.22

Maybe its simply not (yet) added to ondrej surys php repository.

E: ya, its already tracked by these issues: 
- https://github.com/oerdnj/deb.sury.org/issues/1842 
- https://github.com/oerdnj/deb.sury.org/issues/1840

So only way to add APCu compatibility would be by compiling from source as we did with PHP 8.1.
Due to the integration tests of the PSR-6 & PSR-16 decorator, it is not possible to switch to the memory adapter as it does not have persistence layer.

boesing (changes requested):
- [ ] Due to the removal of PHP 7.4, the codestyle needs to be adapted. Might already work by simply running `composer cs-fix`
- [ ] Modify `.laminas-ci/install-apcu-extension-from-source.sh` so that it installs `ext-apcu` for PHP 8.2 from source rather than for PHP 8.1

boesing:
I've added support for PHP 8.2 in this PR:
https://github.com/laminas/laminas-cache-storage-adapter-apcu/pull/45

You can adapt the changes regarding `pre-install.sh` and `install-apcu-extension-[...].sh` from the diff, that should provide APCu compatibility for PHP 8.2.

I wanted to push the changes to your branch but actually this PR does not allow maintainers to push to the branch and thus, I am not able to apply the changes by myself.

fezfez:
> I've added support for PHP 8.2 in this PR: [laminas/laminas-cache-storage-adapter-apcu#45](https://github.com/laminas/laminas-cache-storage-adapter-apcu/pull/45)
> 
> You can adapt the changes regarding `pre-install.sh` and `install-apcu-extension-[...].sh` from the diff, that should provide APCu compatibility for PHP 8.2.
> 
> I wanted to push the changes to your branch but actually this PR does not allow maintainers to push to the branch and thus, I am not able to apply the changes by myself.

Thanks for tips, seem to be ok :)

Ocramius:
@fezfez I opened #238 and am doing some more adjustments on top of your patch, no need to push here :D

Ocramius:
Thanks @fezfez! Merged as part of #238

PR #250, opened 2022-12-23
Adds missing description of usage in a laminas-mvc based application

### LINKED ISSUE #168: Documentation: Add integration examples for `mezzio` and `laminas-mvc`
#### Summary

We do actually lacking examples on how to integrate this library into `laminas-mvc` or `mezzio`.
One of our users found an [old blog post](https://samsonasik.wordpress.com/2013/10/06/zend-framework-2-using-caches-configuration-to-setting-up-cache-services/) written by @samsonasik in 2013 (pointed me on that via Slack).

Luckily, I've chose that configuration style (just the `plugins` config is not compatible with v3.0) and thus that helped the user. 
I would prefer having these examples as part of our documentation.

#### TL;DR

Add examples on how to implement `caches` configuration for the `StorageCacheAbstractServiceFactory` and `cache` configuration for the `StorageCacheFactory`.

froschdesign:


docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+        echo $this->cache->getItem('example') // value;
+        
+        // …
+        
+        return [];
+    }
+}
+```
+
+## Register Controller
```
boesing, line 73: do we need this in the cache component? this might change in upcoming mvc versions but wont change here at the same time. 

Id say lets omit this but up to you as you have more experience regarding documentation

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+    ConfigAbstractFactory::class => [
+        Controller\IndexController::class => [
+            'default-cache',
+        ],
+    ],
+    // …
+];
+</code></pre>
+
+## Using Multipe Caches
```
boesing: typo

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+
+The use more than one cache backend, the factory `Laminas\Cache\Service\StorageCacheAbstractServiceFactory` allows to define multiple cache storages.
+
+Extend the cache configuration in `config/autoload/global.php` and add more cache adapters:
```
boesing: I'd rather suggest a dedicated "cache.global.php" as having plenty stuff in the global quickly becomes messy in bigger projects

boesing (changes requested):

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: If we omit this step, we can throw away the entire description.

> …this might change in upcoming mvc versions…

This may be true, but right now there is no plan or movement on this topic. Therefore, it will continue to be explained that way for now.

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: I'll have to see if creating another file and adding a statement that it will be read automatically will keep the whole description as simple as possible.

boesing: It usually does, the MVC skeleton has a glob-path, same for mezzio:
https://github.com/laminas/laminas-mvc-skeleton/blob/8a4ea5f0583cb518cea62e14ab6f6b8e79a6e96d/config/application.config.php#L24
https://github.com/mezzio/mezzio-skeleton/blob/8f8394a1a4440ab532bfce3fc065d0b01cc29f21/config/config.php#L40

froschdesign: No, I mean we need to explain this fact additionally in this tutorial. Therefore, an already existing file is currently being used.

froschdesign: > I'd rather suggest a dedicated "cache.global.php" as having plenty stuff in the global quickly becomes messy in bigger projects

Normally this note would have to be in the getting started tutorial under a heading like "Optimization" and/or "Best Practice".

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
boesing: Usually, MVC applications do use factories. I'd say it might be enough when we describe on how to configure the cache component so the `StorageInterface` can be retrieved via the `$container` within a factory.

If upstream projects are using `ConfigAbstractFactory`, they know how to configure that since as far as I know that is not the default suggested way of configuring controllers.
I would simply create an example controller class + example factory rather than referring to the ConfigAbstractFactory.

I can only talk for myself but I am working in a MVC application with ~100 controllers and none of them are configured with the `ConfigAbstractFactory`.

froschdesign: At the moment there are no real tools like CLI commands for applications based on laminas-mvc, therefore, everything must be generated by hand. This does not make it easy for beginners.
However, we absolutely need this friendly and simple entrance. And these descriptions under "Application Integrations" must be simple and should use the existing features of the framework / components.

But of course you are right, this again requires a guide that explains the optimization of an application.

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+
+MISSING: **Installation Requirements**
+Make sure that the used storage adapters are installed:
+```bash
+$ composer require laminas/laminas-cache-storage-adapter-memory laminas/laminas-cache-storage-adapter-blackhole
+```
+
+### Change Used Adapter for Controller
+
+To use a different cache adapter for the controller, change the related module configuration and use one of the previously defined names:
```
froschdesign, line 155: @boesing 
With simple use, I mean exactly this step.

Writing another factory by hand only distracts from the topic in this description and is annoying and cumbersome.

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: Some CLI commands can help here to optimise an application by searching and replacing the reflection- and configuration-based factories. This would also be easy to use.

See also: https://github.com/orgs/laminas/projects/17

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: @boesing 
The tutorial has been updated to use a separate configuration file for the caches.

boesing (approved):

boesing:
gnah, damn - merged into 3.9.x 😅

PR #252, opened 2023-01-10
Bump `servicemanager` minimum to 3.18.0

gsteel:
### Description

`laminas-servicemanager:3.18.0` is the first release to support PHP 8.2 - this will also prevent installation of `psr/container:2` which forces installation of `laminas-servicemanager:3.15.0`

Reference: https://github.com/laminas/laminas-cache-storage-adapter-blackhole/pull/57

Also bumps dev deps and resolves new psalm issues


Ocramius (approved):
Thanks @gsteel!

PR #263, opened 2023-07-17
Replace `stella-maris/clock` with `psr/clock`

### LINKED ISSUE #261: Provide native support for PSR-20
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

We do currently require the working group implementation of PSR-20 via `CacheItemPoolDecorator` as at the time of implementing that, PSR-20 was not released.
Since we now have PSR-20 available, we should move to the official PHP-FIG version instead and drop "stellamaris" implementation.

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

With [3.4.0](https://github.com/laminas/laminas-cache/releases/tag/3.4.0), the pre PSR-20 `ClockInterface` was introduced to this package via the dependency of `stella-maris/clock`. With 0.1.7, the project is extending `PSR-20` and thus, can be replaced as long as we conflict with versions lower than `0.1.7`.

This PR updates the code to consume PSR interface rather than working group version.

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,12 +1,10 @@
 {
-    "ignore_php_platform_requirements": {
-        "8.2": true
-    },
     "extensions": [
         "apcu"
     ],
     "ini": [
         "apc.enabled=1",
         "apc.enable_cli=1"
-    ]
+    ],
+    "backwardCompatibilityCheck": true
 }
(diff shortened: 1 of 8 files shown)


Ocramius (approved):
Some tests still require adjustment, but then LGTM :)

boesing:
Handling the BC check issue in https://github.com/laminas/laminas-continuous-integration-action/pull/161

PR #266, opened 2023-07-18
Allow configuration of `AdapterPluginManager` via config file

### LINKED ISSUE #197: RFC: Configure AdapterPluginManager from the config
## RFC

| Proposed Version(s) | x.y.0  |
| BC Break?           | No |

### Goal

Configure `AdapterPluginManager` from the config.

### Background

Every other service manager laminas provides does have a dedicated configuration key (such as `validators`) but the `AdapterPluginManager` component never had such configuration key.

### Considerations

With this feature we would provide users writing custom cache storage adapters a familiar way of configuring `AdapterPluginManager` while not having to write dedicated delegators.

We also do not need to update the various `laminas/laminas-cache-storage-adapter-XXX` repositories. As @boesing mentioned in https://discourse.laminas.dev/t/configure-storageadapterfactory-from-the-config/2918/4 Laminas ensures internally that the vendor adapters are correctly configured using psalm type annotations.

### Proposal(s)

Declare a `storage_adapters` key in the application config and read it in `StorageAdapterPluginManagerFactory` to initialize the `AdapterPluginManager`. 

```php

final class AdapterPluginManager extends AbstractPluginManager
{
    public const CONFIGURATION_KEY = 'storage_adapters';

    /**
     * Do not share by default
     *
     * @var bool
     */
    protected $sharedByDefault = false;

    /** @var string */
    protected $instanceOf = StorageInterface::class;
}

final class StorageAdapterPluginManagerFactory
{
    public function __invoke(ContainerInterface $container): AdapterPluginManager
    {
        $config = $container->get('config')[AdapterPluginManager::CONFIGURATION_KEY] ?? [];
        return new AdapterPluginManager($container, $config);
    }
}
```

rohm1:
PR for #197


docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -37,7 +37,22 @@ return [
 ];
 ```
 
-The factory `Laminas\Cache\Service\StorageCacheAbstractServiceFactory` uses the configuration, searches for the configuration key `caches` and creates the storage adapters using the discovered configuration.
```
boesing: hm, seems that this was removed. Maybe by accident?

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -37,7 +37,22 @@ return [
...
+## Register a storage adapter
```
froschdesign: This is the wrong place for this description as it is unrelated to the rest of the content and is incomplete.
See also: https://github.com/laminas/documentation/issues/8

Please create a new page for this topic and add a complete example that can be fully reproduced by the reader.

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -37,7 +37,22 @@ return [
...
+
+To register additional storage adapters, use either application or module configuration (such as `config/autoload/*.global.php` or `module/Application/config/module.config.php`, respectively), and define the configuration key `storage_adapters`.
```
boesing: Might be helpful to scrape the documentation of other components with this kind of plugin manager configuration. Would love to align the docs somehow, for example:

https://github.com/laminas/laminas-serializer/blob/2.16.x/docs/book/plugin-manager.md

src/Service/StorageAdapterPluginManagerFactory.php
```diff
@@ -9,6 +11,23 @@ final class StorageAdapterPluginManagerFactory
...
+
+        $config = $container->get('config');
+
+        // If we do not have a configuration, nothing more to do
+        if (! isset($config[AdapterPluginManager::CONFIGURATION_KEY]) || ! is_array($config[AdapterPluginManager::CONFIGURATION_KEY])) {
+            return $pluginManager;
+        }
+
+        // Wire service configuration
+        $pluginManager->configure($config[AdapterPluginManager::CONFIGURATION_KEY]);
```
boesing: You might want to store this value into a variable and use inline annotation to provide the `@var` type from https://github.com/laminas/laminas-servicemanager/blob/625f2aa3bc6dd02688b2da5155b3a69870812bda/src/ServiceManager.php#L84

src/Storage/AdapterPluginManager.php
```diff
@@ -16,6 +16,8 @@
  */
 final class AdapterPluginManager extends AbstractPluginManager
 {
+    public const CONFIGURATION_KEY = 'storage_adapters';
```
boesing: Can we move this constant into the `ConfigProvider` instead?

boesing (changes requested):

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: Here is another example: https://docs.laminas.dev/laminas-hydrator/v4/plugin-managers/

But it should be explicitly described, as in this example: https://github.com/laminas/laminas-form/pull/196/files#diff-3a14a07ea6ab87492ab3919a659959973bb061f53b2e3a84395a9e0f4c2dae2b

docs/bookdown.json
```diff
@@ -3,6 +3,7 @@
     "target": "html/",
     "content": [
         "book/laminas.cache.storage.adapter.md",
+        "book/laminas.cache.storage.adapter-plugin-manager.md",
```
froschdesign: [All `bookdown.json`](https://github.com/search?q=org%3Alaminas%20bookdown.json&type=code) files can be killed because we done not use [Bookdown](http://bookdown.io).

docs/book/v3/storage/adapter-plugin-manager.md
```diff
@@ -0,0 +1,28 @@
+# Adapter Plugin Manager
+
+The `AdapterPluginManager` extends the laminas-servicemanager `AbstractPluginManager`, and has the following behaviors:
+
+- It will only return `Laminas\Cache\Storage\StorageInterface` instances.
+- All services are not shared by default; a new instance will be created each time you call `get()`.
```
froschdesign: An introduction would be good here: Why should a user use this plug-in manager, what are the benefits and what are the advantages?

Example: https://github.com/laminas/laminas-form/pull/196/files#diff-3a6ef754f12eb3c6e8ccb63159f571bf326bf50fdf9304706dc9188e3af06f77R1-R20

docs/book/v3/storage/adapter-plugin-manager.md
```diff
@@ -0,0 +1,28 @@
...
+```php
+// config/autoload/storage_adapters.global.php
+return [
+    'storage_adapters' => [
+        'factories' => [
+            \App\MyCustomStorageAdapter::class => \App\Container\MyCustomStorageAdapterFactory::class,
+        ],
+    ],
+];
+```
```
froschdesign: What is the goal of this section? To explain the `StorageAdapterPluginManagerFactory` or how a user can add his own storage adapter?

docs/book/v3/storage/adapter-plugin-manager.md
```diff
@@ -0,0 +1,28 @@
...
+- It will only return `Laminas\Cache\Storage\StorageInterface` instances.
+- All services are not shared by default; a new instance will be created each time you call `get()`.
+
+## Factory
+
+`Laminas\Cache\Storage\AdapterPluginManager` is mapped to the factory `Laminas\Cache\Service\StorageAdapterPluginManagerFactory` when wired to the dependency injection container.
+
+The factory will be automatically registered when loading/installing the `Laminas\Cache` module in `laminas-mvc` and/or loading/installing the `ConfigProvider` into a Mezzio application.
+
+Since version 3.11.0, the factory will look for the `config` service, and use the `storage_adapters` configuration key to seed it with additional services.
```
froschdesign: The version hints must be highlighted:

## Version Hints

If a new feature was added with a minor release then a version hint must be added to the related section of the documentation page.
The hint must be added after the headline of the new feature.

An admonition must be used for a version hint.

### Syntax Example

``````
INFO: **New Feature**
Available since version 3.11.0
``````

froschdesign (changes requested):

docs/book/v3/storage/adapter-plugin-manager.md
boesing: @froschdesign We do already have this plugin manager and this PR just adds another way to provide cache adapters which are not provided by laminas.

I personally do not see a reason why userland has a need for this plugin manager (besides the fact projects provide own cache adapters for niche backends). Those who already do provide a cache adapters do have to register these via a delegator factory and thus do already register these delegators to the `service_manager`/`dependencies` key so that their adapter is being added to the plugin manager. This PR just provides a dedicated cache key so that there is no runtime delegator to be executed on each request.

Not sure why we should provide a "why should a user use this plugin manager", esp. since userland should use `StorageAdapterFactoryInterface` instead which is already documented here: 
https://github.com/laminas/laminas-cache/blob/3.11.x/docs/book/v3/storage/adapter.md#quick-start

Does that suit our needs so far or do we have to add your example?

docs/book/v3/storage/adapter-plugin-manager.md
boesing: The latter, it is about how to add cache adapters to the plugin manager.
We do have the exact same documentation in the serializer component and thus I'd recommended @rohm1 to use that:
https://github.com/laminas/laminas-serializer/blob/2.16.x/docs/book/plugin-manager.md

docs/book/v3/storage/adapter-plugin-manager.md
boesing: Can this be added to the very first line of this markdown or where do I have to put that?

docs/book/v3/storage/adapter-plugin-manager.md
froschdesign: @boesing 
I am not judging here whether the plugin manager is needed or not, but the newly added documentation. And if the plugin manager is to be described, then there must also be a justification of what advantages it offers and why the user should use it.

If you say that the plugin manager is not needed, then this documentation is unnecessary too.

boesing: The plugin manager is needed once you create your own storage backend. and in this case, you can now - with this Patch - provide factories for your own written storage adapters in the same way as in every other component => via config. 

Thats all it provides, reading config and passing it to the plugin manager.

I dont see why we have to add more documentation than how to pass a valid config to the plugin manager factory. But maybe I miss your point?

froschdesign: Let's approach it differently: Should the user act directly with the plugin manager at any point, e.g. pass it to a controller and then fetch an adapter from the plugin manager there?
If not, then we don't need detailed documentation, but if the user deals directly with the plugin manager - not only via the configuration - then we need a detailed documentation for it.

boesing: The main suggestion by `laminas-cache` is to not use the plugin manager but to use `StorageAdapterFactoryInterface` instead, which is a dedicated service to create a storage adapter based on configuration.

We do use the https://github.com/laminas/laminas-cache/blob/3.11.x/src/Service/StorageCacheAbstractServiceFactory.php along with `caches` array config to fetch adapters from the config rather than using neither a service or the plugin manager.

froschdesign: > The main suggestion by `laminas-cache` is to not use the plugin manager…

Then goes this current documentation in the wrong direction and we need something like "Configuration" under the section "Application Integration". Here can be described the new configuration key `storage_adapters`.
The plugin manager itself then only appears in a side sentence, with the reference that it is an implementation of the plugin manager based on laminas-servicemanager.

boesing (approved):
LGTM, thanks @rohm1

Issue #130, opened 2021-07-11
[RFC]: Move `StorageInterface::getMetadata` to `MetadataCapableInterface`


boesing:
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 4.00  |
| BC Break?           | Yes |

### Goal

Having those adapters which do support metadata must explicitly implement that interface rather than assuming that all adapters will have that capability.

### Background

Just a handful of adapters do support Metadata. Almost all adapters have different metadatas aswell. With having a dedicated interface which also have a template to annotate the generic Metadata instance with getters of these metadatas will provide properly typed informations long-term.

Given the example of the filesystem adapter which can provided filesystem-specific metadatas (`mtime`, `atime`, ...), almost all metadatas are already "explicitly" used. So upstream projects will probably implement `Filesystem` directly rather than using `StorageInterface`. 
If they do, there might be a point where they use `assert($storage instanceOf Filesystem);` because otherwise, that wont make any sense when using `getMetadata` while also accessing the `atime` array key.

With having that new interface in addition with the generic annotation to mark the `Metadata` instance, we can provide a `FilesystemMetadata` instance which then provides type-safe getters for `mtime` and `atime`.

So in the end, all adapters have to provide their own `metadata` implementation.

### Considerations

As already mentioned, I assume that those users which do make use of the `StorageInterface::getMetadata` do this, because they know the underlying storage implementation.
Currently usage would be something like:

```php
$storage = new Filesystem();

$metadata = $storage->getMetadata('key1');
assert($metadata !== false);

$cacheCreatedOn = $metadata['atime'];
```

This would never work with the `Redis` or `Memcached` implementation. So in the end, we will provide proper `FilesystemMetadata` instance with properly typed methods to provide metadata informations of existing cache keys. Migration to the new functionality would also quite easy because instead of accessing array keys, users only have to access the provided getters/setters.

### Proposal(s)

The `MetadataCapableInterface` would have been asserted by either using `instanceof` or by directly consuming the adapter rather than the interface.
I guess, thats already the case in those projects which do use the `StorageInterface::getMedata`

### Appendix

The new `MetadataCapableInterface` might look like this:

```php

/**
 * @template TMetadata of object
 */
interface MetadataCapableInterface
{
	/**
	 * @psalm-return TMetadata
	 */
	public function getMetadata(string $key): ?object;
}
```

A possible implementation in the `Filesystem` adapter could look like this:

```php
[...]

PR #272, opened 2023-08-26
Remove dependency on real storage adapters and therefore benchmarks

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

In the past, we had several problems providing proper support, i.e. for APCu on new PHP versions, etc.
IMHO, there were too many integration tests inside this repository which was kinda helpful when developing within the PSR decorators but on the other side, introduced circular dependencies which was especially annoying when working on new major versions (like as we do right now).

---

This also removes benchmarks which are so available in each adapter repository and thus were redundant.
Even tho, we are now unable to directly see impact on performance when changing `AbstractAdapter`, the adapters themselves keep track on performance due to renovate updating locks permanently.
    
Longterm, the `AbstractAdapter` will most likely be removed in favor of cache backend specific features explicitly implemented in the storage adapters directly (which is already done in some of the adapters anyways). The `StorageInterface` will extend the PSR interfaces and requires the adapters to serve these interfaces in their own way. Therefore, removing benchmarks might be a good thing since the `StorageInterface` is meant to be feature complete.
    
With #130, some methods might be removed from the `StorageInterface` but extending that interface is not planned.


boesing:
@Ocramius I'd love to get your two cents on this. Esp. the benchmark and integration test stuff.

PR #273, opened 2023-08-26
Introduce `MetadataCapableInterface`

### LINKED ISSUE #130: [RFC]: Move `StorageInterface::getMetadata` to `MetadataCapableInterface`

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | https://github.com/laminas/laminas-cache/pull/274/commits/51433fb3e04ceef4ff2dc79f98eb8ce8d4f03fdf
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes

## New Features

- Every adapter which supports `metadata` now implements `MetadataCapableInterface` and provides a dedicated object containing all the metadata values it supports.

## Breaking Changes

- `AbstractAdapter` and `StorageInterface` are not aware of the methods `getMetadata` anymore. These were moved to the new `MetadataCapableInterface`
- `Capabilities` do not provide `supportedMetadata` anymore. The supported metadata is tied to the used storage adapter and thus, was already requiring projects to explicitly know the exact implementation of the cache backend in case of using these metadatas anyway
- `KeyListIterator` and the corresponding `IteratorInterface` does not provide the `mode` `CURRENT_AS_METADATA` anymore 

----

Fixes #130


boesing:
Contains changes of #272

Feature implementation: https://github.com/laminas/laminas-cache/pull/273/commits/6a08ab0b78d0c4b7bfc9e441a2fea167b3c1a5d4

test/Storage/Adapter/AbstractMetadataCapableAdapterTest.php
boesing, line 1: mostly copy & paste from `AbstractAdapterTest`

PR #196, opened 2022-05-19
Add suppport for `psr/cache` v2 and v3

jguittard:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | not yet
| New Feature   | yes

### Description

This adds support for `psr/cache` v2 and v3.

PR #275, opened 2023-08-27
Introduce support for PSR-6 and PSR-16 v2 & v3

### LINKED ISSUE #133: [RFC]: Support for PSR-6 & PSR-16 v3
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no

#### Summary

PSR-6 and PSR-16 has v2 & v3 available. Since I dont see the point of supporting v2 **AND** v3, we should be fine with supporting v1 and v3.
We could use `composer/deprecated-package-versions` along with an autoload check for the `psr/cache` version to use `class_alias` to provide support for both v1 & v3 of `psr/cache` and `psr/simple-cache`.

Same compatibility layer is used in [laminas/laminas-cli](https://github.com/laminas/laminas-cli/blob/ece0c39671097ef7b8e4123b20e4f07f9327b130/src/autoload.php#L14) to support both `symfony/console` v4 & v5.

We could also use `composer-runtime-api` requirement but I am not yet fully convinced that **anyone** is able to use composer in v2 but I am open for this change, if the majority is fine with having a hard requirement to the `composer-runtime-api`.

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | no
| BC Break      | yes
| New Feature   | no

### Description

This introduces support for PSR-6 and PSR-16 v2 & v3.

Supersedes #205, #196, #214
Fixes #133

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,3 +1,13 @@
 {
-    "backwardCompatibilityCheck": true
+    "backwardCompatibilityCheck": true,
+    "additional_checks": [
+        {
+            "name": "Unit tests with psr/cache v2",
+            "job": {
+                "php": "*",
+                "dependencies": "latest",
+                "command": "composer require psr/cache:^2.0 && vendor/bin/phpunit"
+            }
+        }
+    ]
 }
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -5,8 +5,17 @@
             "name": "Unit tests with psr/cache v2",
             "job": {
                 "php": "*",
-                "dependencies": "latest",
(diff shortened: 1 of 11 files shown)


boesing:
This includes changes of #272 
Feature implemented in https://github.com/laminas/laminas-cache/pull/275/commits/21503c13a36ede23d53d56abbfd00c7a6f176edd

MatthiasKuehneEllerhold:
@boesing Any news on releasing this change?

Ocramius:
@MatthiasKuehneEllerhold see the milestone

PR #286, opened 2024-01-17
Infinite loop when fetching some cache-related services via container in projects without registered `config` service

### LINKED ISSUE #284: ServiceManager::get results in endless loop since 3.11.0
### BC Break Report

|    Q        |   A
|------------ | ------
| Version     | 3.11.0, 3.12.0
| Last known good     | 3.10.1

#### Summary

`ServiceManager::get` calls result in an endless loop unless the "config" service has been configured

#### Previous behavior

Works as expected

#### Current behavior

Endless loop

#### How to reproduce

```php
<?php

declare(strict_types=1);

use Laminas\Cache\ConfigProvider;
use Laminas\Cache\Service\StorageAdapterFactoryInterface;
use Laminas\ConfigAggregator\ConfigAggregator;
use Laminas\ServiceManager\ServiceManager;

require_once __DIR__ . '/vendor/autoload.php';

$configAggregator = new ConfigAggregator([ConfigProvider::class]);
$config = $configAggregator->getMergedConfig();
$container = new ServiceManager($config['dependencies']);

// Without this, the get() call below results in an endless loop with laminas/laminas-cache >= 3.11.0
//$container->setService('config', $config);

$container->get(StorageAdapterFactoryInterface::class);
```

InvisibleSmiley:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Add regression test for issue #284


boesing (approved):
LGTM :trollface:

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -37,7 +37,12 @@ return [
 ];
 ```
 
-The factory `Laminas\Cache\Service\StorageCacheAbstractServiceFactory` uses the configuration, searches for the configuration key `caches` and creates the storage adapters using the discovered configuration.
+The factory `Laminas\Cache\Service\StorageCacheAbstractServiceFactory` uses the configuration, searches for the configuration key `caches` and creates the storage adapters using the discovered configuration. 
+
+> ### Cache named `config` is not possible
+>
+> A cache named `config` is not possible due to internal service conflicts with MVC configuration.
+> The service named `config` is reserved for he projects configuration and thus cannot be used with the `caches` configuration.
```
InvisibleSmiley: Typo "the"

gsteel:
```suggestion
> The service named `config` is reserved for project configuration and thus cannot be used with the `caches` configuration.
```

src/Service/StorageCacheAbstractServiceFactory.php
```diff
@@ -14,7 +14,8 @@
  */
 class StorageCacheAbstractServiceFactory implements AbstractFactoryInterface
 {
-    public const CACHES_CONFIGURATION_KEY = 'caches';
+    public const CACHES_CONFIGURATION_KEY       = 'caches';
+    private const PRESERVED_CONFIG_SERVICE_NAME = 'config';
```
gsteel:
```suggestion
    private const RESERVED_CONFIG_SERVICE_NAME = 'config';
```

src/Service/StorageCacheAbstractServiceFactory.php
```diff
@@ -32,6 +33,10 @@ class StorageCacheAbstractServiceFactory implements AbstractFactoryInterface
      */
     public function canCreate(ContainerInterface $container, $requestedName)
     {
+        if ($requestedName === self::PRESERVED_CONFIG_SERVICE_NAME) {
```
gsteel:
```suggestion
        if ($requestedName === self::RESERVED_CONFIG_SERVICE_NAME) {
```

test/Service/ConfigProviderIntegrationTest.php
```diff
@@ -0,0 +1,92 @@
...
+use function is_array;
+use function is_string;
+
+final class ConfigProviderIntegrationTest extends TestCase
+{
+    private ContainerInterface $container;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
```
InvisibleSmiley, line 26: Superfluous, parent is empty

gsteel (commented):
Spotted a typo in the docs, and also the constant name `PRESERVED_` should be `RESERVED_` right?

test/Service/ConfigProviderIntegrationTest.php
```diff
@@ -0,0 +1,92 @@
...
+            yield $serviceName => [$serviceName];
+        }
+    }
+
+    /**
+     * @psalm-assert array<string,mixed> $iterable
+     */
+    private static function assertMappedWithStrings(mixed $iterable): void
+    {
+        if (! is_array($iterable)) {
```
InvisibleSmiley: Should use `is_iterable` here instead of `is_array`, otherwise the parameter name is a misnomer.
OTOH, resetting a Generator after traversing it (in order to determine the key types) is not always possible.
So I think the correct thing to do would to just not pretend that we accept any iterable here and call it what it is instead: `$array`

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: Admonitions/callouts should be used:

```markdown
WARNING: **Cache Named `config` Is Not Possible**
A cache named `config` is not possible due to internal service conflicts with MVC configuration.
The service named `config` is reserved for he projects configuration and thus cannot be used with the `caches` configuration.
```

https://oprypin.github.io/markdown-callouts/index.html#usage

boesing: Seriously, this whole documentation stuff has so many features. Its more like a dedicated scripting language 😂

froschdesign: Markdown is unsuitable for technical documentations, the syntax is not sufficient and extensions are crap. :stuck_out_tongue_winking_eye:

src/Service/StorageCacheAbstractServiceFactory.php
```diff
@@ -14,7 +14,8 @@
  */
 class StorageCacheAbstractServiceFactory implements AbstractFactoryInterface
 {
-    public const CACHES_CONFIGURATION_KEY = 'caches';
+    public const CACHES_CONFIGURATION_KEY      = 'caches';
+    private const RESERVED_CONFIG_SERVICE_NAME = 'config';
```
Xerkus, line 18: Why introduce constant? Lines 76 and 81 are not using it.

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -37,7 +37,11 @@ return [
 ];
 ```
 
-The factory `Laminas\Cache\Service\StorageCacheAbstractServiceFactory` uses the configuration, searches for the configuration key `caches` and creates the storage adapters using the discovered configuration.
+The factory `Laminas\Cache\Service\StorageCacheAbstractServiceFactory` uses the configuration, searches for the configuration key `caches` and creates the storage adapters using the discovered configuration. 
+
+WARNING: **Cache Named `config` Is Not Possible**
+A cache named `config` is not possible due to internal service conflicts with MVC configuration.
+The service named `config` is reserved for project configuration and thus cannot be used with the `caches` configuration.
```
Xerkus, line 44: Cache name must be unique identifier in container. Not just config but any conflicting name is not possible.

boesing: yah but thats usually up to the user while it is not for `config` as we now do have explicit check within the abstract static factory.
So I think this might be good enough for now. We still should tackle this specific issue within `laminas-servicemanager`.

src/Service/StorageCacheAbstractServiceFactory.php
boesing: because its way more readable than just having a random string somewhere.
The `get` does not need to check for the requested name as it shouldn't get passed due to the `canCreate` method returning false (implicit check).

I did a quick lookup with `phpdbg` and from what I can see, having this constant does not even produce more opcodes than using string here.

So IMHO, due to readability and same opcode amount, it should be fine having this around.

PR #289, opened 2024-01-19
Remove virtual package dependency of `laminas/laminas-cache-storage-implementation`

### LINKED ISSUE #278: Remove virtual package requirement `laminas/laminas-cache-storage-implementation`
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

When working on `laminas-cache-storage-adapter-benchmark` and `laminas-cache-storage-adapter-test`, having to fulfill the requirement of `laminas/laminas-cache-storage-implementation` by requiring one of the adapters introduces a bunch of pain.

Whenever there are new major versions to `laminas-cache`, this becomes a huge pain since first, benchmark and test packages have to support the new major before the adapters itself can get updated and then the actual implementations can bump to the new major (including tests, etc.).

We initially added `laminas/laminas-cache-storage-implementation` to ensure that there are no projects out there to "upgrade" without having any implementation in-place. With v4, we could now remove this requirement.
Any setup guide is already pointing the users to require the specific cache backend + laminas-cache itself.

I already proposed a change to composer itself to be able to kinda "ignore" virtual packages for non-project packages in https://github.com/composer/composer/issues/11671 but that seems to be against what composer expects how dependency inversion works. 
I still think it would be helpful if composer is actually ignoring virtual package requirements in "library" packages but I don't think that this will happen any time soon and thus, I propose to remove `laminas/laminas-cache-storage-implementation` in v4.0.0.

Any thoughts about this?

boesing:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

`laminas/laminas-cache` can be required by libraries, etc. without having to care about specific implementations. To avoid those libraries to have to require at least one specific implementation during development, the virtual package dependency is being removed. Virtual package dependencies should be moved into projects rather than libraries.

Closes #278


Ocramius (approved):
👍

PR #294, opened 2024-02-28
Removal of `increment` and `decrement` functionality

### LINKED ISSUE #117: [RFC]: Removal of increment/decrement functionality
## RFC

| Proposed Version(s) | 4.0.0  |
| BC Break?           | Yes |

### Goal

It seems that some storage backends (`redis` for example) does not preserve type-safety when storing values.
Incrementing and decrementing of cache items is therefore mostly done by this library by reading the value, Incrementing/decrementing the value and then it is written back to the storage.
This could also be done in userland code and/or by a decorator and thus does not need a requirement for a storage anymore.

### Background

Recently, there were 2 bugs discovered when using the serializer plugin in combination with decrement/increment which were  undetected for around 9 years.

### Considerations

Providing a decorator which consumes the storage along with the key & amount could provide the same functionality without having every storage do handle it by itself.

### Proposal(s)

Thus said, I would mark that whole functionality deprecated in 3.0 and remove it in 4.0

Decorator to increment/decrement could be created in 3.0 if needed. The whole functionality could be done in userland without having to use a decorator at all and thus the decorator could also be just an example in the docs.

### Appendix

Bugs mentioned in this RFC which are related to this functionality:
#116
#115

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

As per #117, the increment and decrement functionality (at least decrement) leads to issues with several backends. This commit removes that functionality at all as it should be either part of specific adapters instead (via dedicated interface) or as part of project specific code. This cache component can not guarantee increment/decrement for all adapters out there (especially for adapters which do not support this feature natively to prevent race conditions).


boesing:
BC compatibility check is expected, this PR introduces a BC break on purpose.

----

@Ocramius does it make sense to add a way that the BC check is passing once `BC Break` label is added to a PR?
I mean, I guess its fine that the check is red here as it actually does introduce a BC break but it still feels kinda "wrong" as it is expected. So this is just an idea I'd like to bounce.

boesing:
- [x] Remove event listeners listening to `incrementItem`, `incrementItems`, `decrementItem` and `decrementItems` events

Ocramius:
> BC compatibility check is expected, this PR introduces a BC break on purpose.
> 
> @Ocramius does it make sense to add a way that the BC check is passing once `BC Break` label is added to a PR? I mean, I guess its fine that the check is red here as it actually does introduce a BC break but it still feels kinda "wrong" as it is expected. So this is just an idea I'd like to bounce.

IMO NO: can just decide to merge with red, which is a clear/visible flag

PR #295, opened 2024-02-29
Introduce native type-hints and pass-by-reference values

### LINKED ISSUE #14: reduce arguments by reference
This PR reduces arguments by reference often used on internal methods.

It was previously done as performance improvement but in fact it decreases performance as PHP have to create a new zval and in some cases it needs to copy the value before PHP-7 as described here: http://nikic.github.io/2015/05/05/Internal-value-representation-in-PHP-7-part-1.html

Additionally arguments by reference make the code more error prone and hard to read.

---

Originally posted by @marc-mabe at https://github.com/zendframework/zend-cache/pull/8

### LINKED ISSUE #15: Return values of some `AbstractAdapter` methods do not reflect the annotated return type

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

This introduces almost all native type-hints. Those which are missing are either related to `laminas-eventmanager` or `laminas-servicemanager`. The latter will be updated in one of the next PRs.

This fixes #15


boesing:
This contains all changes of #294 as well.

src/Pattern/CallbackCache.php
```diff
@@ -23,13 +23,12 @@ class CallbackCache extends AbstractStorageCapablePattern
     /**
      * Call the specified callback or get the result from cache
      *
-     * @param  callable   $callback  A valid callback
+     * @param  callable|callable-string   $callback  A valid callback
```
Ocramius: The need for this declaration is really strange: IMO wrong too.

Should just be `callable`, letting the PHP engine check it too :thinking:

src/Pattern/ObjectCache.php
```diff
@@ -284,11 +282,35 @@ public function __toString(): string
...
+    private function generateArgumentsKey(array $args): string
+    {
+        if ($args === []) {
+            return '';
+        }
+
+        ErrorHandler::start();
+        try {
+            $serializedArgs = serialize(array_values($args));
+        } catch (\Exception $e) {
```
Ocramius: `\Throwable`

Ocramius (commented):
Only managed to look at very superficial level: test changes OK, only change that is a bit weird is `callable|string` types, which I think is wrong :thinking:

boesing:
Will double check that again. Problem is that it was "callable" beforehand and I am unsure if it allows strings pointing to functions.

Ocramius:
> Will double check that again. Problem is that it was "callable" beforehand and I am unsure if it allows strings pointing to functions.

see https://3v4l.org/qCIk0

```php
<?php

function useCallable($input, callable $cb): mixed {
    return $cb($input);
}

var_dump(useCallable('123', 'intval'));
var_dump(useCallable('', 'intval'));
var_dump(useCallable('foo', 'intval'));
var_dump(useCallable(123.456, 'intval'));
```

PR #274, opened 2023-08-26
Documentation for v4.0.0 including migration guide

boesing:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

This introduces initial `v4` documentation including an initial migration guide. This is work in progress and thus might change with other PRs in the future.

Adapter modifications can be reviewed here:
https://github.com/laminas/laminas-cache/pull/274/commits/51433fb3e04ceef4ff2dc79f98eb8ce8d4f03fdf

PR #296, opened 2024-03-03
Support for `laminas/laminas-servicemanager` v4

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | no

### Description

This also introduces support for `laminas/laminas-serializer` v3 and conflicts with `laminas-serializer` <3.0.

- [x] Add BC break regarding `PluginOptions#getSerializer` does not create a serializer anymore if a `string` was passed beforehand (added to #274)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
         "laminas/laminas-eventmanager": "^3.4",
-        "laminas/laminas-servicemanager": "^3.21",
+        "laminas/laminas-servicemanager": "^4.0",
         "laminas/laminas-stdlib": "^3.6",
         "psr/cache": "^2.0 || ^3.0",
         "psr/simple-cache": "^2.0 || ^3.0",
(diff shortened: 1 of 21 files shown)


src/Storage/PluginManager.php
```diff
@@ -14,57 +19,57 @@
...
-        Plugin\IgnoreUserAbort::class      => InvokableFactory::class,
-        Plugin\OptimizeByFactor::class     => InvokableFactory::class,
-        Plugin\Serializer::class           => InvokableFactory::class,
+    private const CONFIGURATION = [
+        'aliases'   => [
+            'clear_expired_by_factor' => Plugin\ClearExpiredByFactor::class,
+            'clearexpiredbyfactor'    => Plugin\ClearExpiredByFactor::class,
+            'clearExpiredByFactor'    => Plugin\ClearExpiredByFactor::class,
+            'ClearExpiredByFactor'    => Plugin\ClearExpiredByFactor::class,
+            'exception_handler'       => Plugin\ExceptionHandler::class,
```
gsteel, line 33: Is there any value in keeping all of these aliases alive? Would it not be a good time to get rid of them and force users to reference the required implementation by FQCN?

gsteel (approved):

src/Storage/PluginManager.php
Ocramius: Just BC

composer.json
```diff
@@ -47,12 +47,14 @@
         "laminas/laminas-cli": "^1.7",
         "laminas/laminas-coding-standard": "~2.5.0",
         "laminas/laminas-config-aggregator": "^1.13",
-        "laminas/laminas-serializer": "^2.14",
+        "laminas/laminas-serializer": "3.0.x-dev",
+        "laminas/laminas-servicemanager": "^4.0.0-rc3",
```
samsonasik: require-dev seems not needed for `laminas/laminas-servicemanager` as it already in `require` above, also, `4.0.0` already released.

boesing: yah this PR is older than v4.0.0 release. Will update at some point.

src/Storage/PluginManager.php
boesing: I was also thinkin of this but besides the fact that it would provide a benefit to projects with psalm, I think its okayish to keep them. Might be something to consider in next major but I would do that with a better way to configure adapters in total, not just enforcing FQCN here.

PR #310, opened 2024-06-14
Enhance type strictness and add `final` to a bunch of classes

boesing:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

This adds a bunch of `final` statements to classes which were not expect to be inherited.


gsteel (approved):
LGTM 👍

PR #319, opened 2024-06-21
Add missing template param for `StorageInterface`

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | soft
| New Feature   | no

### Description

Since all laminas cache adapters are extending `AbstractStorage`, there was already a template `TOptions` available in that abstract class. Sadly, I forgot to add that to the `StorageInterface` which is problematic when it comes to interface usage.

This patch adds options template as a bugfix since we do not change any code (besides providing the options generic).

### DIFF
--- a/src/Storage/Adapter/AbstractAdapter.php
+++ b/src/Storage/Adapter/AbstractAdapter.php
@@ -31,6 +31,7 @@ use function sprintf;
 
 /**
  * @template TOptions of AdapterOptions
+ * @template-implements StorageInterface<TOptions>
  */
 abstract class AbstractAdapter implements StorageInterface, PluginAwareInterface
 {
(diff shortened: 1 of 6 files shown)


gsteel (approved):
👍 - Out of interest, why do the templates not need to be co-variant?

boesing:
> Out of interest, why do the templates not need to be co-variant?

I have absolutely no clue when it has to be and when not.
I've used covariant once but I cannot remember when. AFAIR covariant is only used if you want to mix stuff such as "pets" where "dog" and "cat" has to be in there.
Storage adapters only provide a single generic and thus covariant is not needed as from how I understand.

But I am not that deep into that stuff as I never really cared 😅

PR #350, opened 2024-11-25
Fix docs linting issues

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes


gsteel:
@froschdesign Can you advise what to do about `/docs/index.md` - can it be deleted?

froschdesign:
@gsteel 
> …can it be deleted?

Unfortunately not, because MkDocs produce an error if the file is listed in the navigation but not in the file system. And the navigation entry is needed, otherwise the homepage is missing.

froschdesign:
And here we have special case because the content of `index.md` is used for the homepage instead of auto-generated:

https://github.com/laminas/laminas-cache/blob/f25e67a99311bbcfeeda6f8167419f3518a1e691/docs/book/index.md?plain=1#L9-L15

https://docs.laminas.dev/laminas-cache/#learn

froschdesign (approved):

gsteel:
I've just disabled linting for index.md - that OK?

PR #357, opened 2025-01-10
Allow PHP 8.4 in laminas-cache 3.x

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

This MR adds support for PHP 8.4 to laminas-cache v3. Adding support to the legacy version is essential, as laminas-cache v4 requires servicemanager v4 and many users cannot yet migrate to servicemanager v4, as quite a few packages in the MVC ecosystem yet lack support for servicemanager v4. Hence, it makes much sense to add support for PHP 8.4 to laminas-cache v3, given that this is pretty straight foward.

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -8,6 +8,6 @@
     ],
     "backwardCompatibilityCheck": true,
     "ignore_php_platform_requirements": {
-        "8.3": true
+        "8.4": true
     }
 }
--- a/.laminas-ci/pre-install.sh
+++ b/.laminas-ci/pre-install.sh
@@ -2,17 +2,10 @@
 
 WORKING_DIRECTORY=$2
 JOB=$3
-PHP_VERSION=$(php -nr "echo PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION;")
+PHP_VERSION=$(echo "${JOB}" | jq -r '.php')
 
-if [ ! -z "$GITHUB_BASE_REF" ] && [[ "$GITHUB_BASE_REF" =~ ^[0-9]+\.[0-9] ]]; then
-  readarray -td. TARGET_BRANCH_VERSION_PARTS <<<"${GITHUB_BASE_REF}.";
-  unset 'TARGET_BRANCH_VERSION_PARTS[-1]';
-  declare -a TARGET_BRANCH_VERSION_PARTS
-  MAJOR_OF_TARGET_BRANCH=${TARGET_BRANCH_VERSION_PARTS[0]}
-  MINOR_OF_TARGET_BRANCH=${TARGET_BRANCH_VERSION_PARTS[1]}
-
-  export COMPOSER_ROOT_VERISON="${MAJOR_OF_TARGET_BRANCH}.${MINOR_OF_TARGET_BRANCH}.99"
-  echo "Exported COMPOSER_ROOT_VERISON as ${COMPOSER_ROOT_VERISON}"
-fi
+# Temporary workaround for cyclic dependencies - must be adjusted in next major branch!
+echo "Branch as 3.99.x in Pre-Install"
+git checkout -b 3.99.x
 
 ${WORKING_DIRECTORY}/.laminas-ci/install-apcu-extension-via-pecl.sh "${PHP_VERSION}" || exit 1
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         }
     },
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
+        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
         "laminas/laminas-cache-storage-implementation": "1.0",
         "laminas/laminas-eventmanager": "^3.4",
         "laminas/laminas-servicemanager": "^3.21",
--- a/composer.lock
+++ b/composer.lock
@@ -4,44 +4,45 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e30d07ce4a69432401e9676e46993cd1",
+    "content-hash": "60d0b4f8ae10814cda0d9bbe6bbdfad4",
     "packages": [
         {
-            "name": "laminas/laminas-cache-storage-adapter-memory",
-            "version": "2.2.0",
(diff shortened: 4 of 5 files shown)


driehle:
@froschdesign 

This is a follow-up on doctrine/DoctrineORMModule#767, on how PHP 8.4 support could be added to laminas-cache 3.x series. Technically it works, there are, however, some issues:

Due to the architecture of laminas-cache, there is [a dependency on the root project](https://getcomposer.org/doc/articles/troubleshooting.md#dependencies-on-the-root-package). Therefore, Composer needs to inflect the version of the root package and, obviously, fails to do so. The initial error in CI was:

```
    - laminas/laminas-cache-storage-adapter-filesystem 2.4.1 requires laminas/laminas-cache ^3.10.0 
      -> found laminas/laminas-cache[4.2.x-dev] but it does not match the constraint. 
           See https://getcomposer.org/dep-on-root for details and assistance.
```

So it seems to inflect 4.2.x-dev, though 3.13.x-dev is what the value should be. First of all, there was typo in the `pre-install.sh` script, where `COMPOSER_ROOT_VERISON` was set. However, I think this script does not have any effect at all, at least fixing the typo doesn't change anything. I think this makes sense, since variables exported in Bash are lost when the script exits. So when the actuall Composer call is made, the variable will not be present anymore.

For now, I have set the version in `composer.json`, which is a bit hacky and ugly. A clean solution would probably be to integrate the code from the `pre-install.sh` script somwhere in the laminas-ci-matrix-action repository, so that `COMPOSER_ROOT_VERSION` can actually be set based on the target branch of the MR when Composer is called.

With the version pinned, most CI checks work. The docs seem to be failed, but I'd call this unrelated. The Psalm baseline probably needs to be updated, which I haven't done yet and the backward compatibility check, I have no idea what this is about. 🤷

driehle:
I tried setting `COMPOSER_ROOT_VERSION` as @boesing suggested in laminas/laminas-continuous-integration-action#35, i.e. the GitHub way by writing to `$GITHUB_ENV`. However, this doesn't work and after thinking about it, in indeed does make sense that it does not work: Both `composer install` and the following `phpunit` call are executed in the same step, as there is only one CI step, which is the `laminas/laminas-continuous-integration-action` handling all the stuff in one single step. And according to [GitHub docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#passing-values-between-steps-and-jobs-in-a-workflow) the `GITHUB_ENV` is only for passing variables between different steps.

Therefore, I am surprised how the code in https://github.com/laminas/laminas-cache-storage-adapter-blackhole/blob/fe7bb4045fae6a71f8179744829aab6ddf47521e/.laminas-ci/composer-root-version.sh was supposed to work. In the same way, I am surprised how the code in this repository, using `export COMPOSER_ROOT_VERSION`, should ever have worked. Variables set in bash scripts are lost, once the script exits. This would only be possible, if the script is sourced, but that is not happening in https://github.com/laminas/laminas-continuous-integration-action/blob/1.41.x/entrypoint.sh#L176.

driehle:
@gsteel @samsonasik Any suggestions on how to proceed here?

gsteel:
Hi @driehle - In `laminas-view` we checkout a new branch prior to running CI: https://github.com/laminas/laminas-view/blob/2.37.x/.laminas-ci/pre-install.sh

driehle:
Great idea, that works like a charm! 👏 

Now all tests except for the docs linting are passing. I won't fix the docs linting though, as Laminas only renders docs from the latest branch (i.e. currently `4.2.x`) and docs from legacy branches, like `3.13.x` here, are not used anyways. Do you agree?

src/Storage/Capabilities.php
```diff
@@ -162,6 +162,7 @@ public function getAdapter()
     /**
      * Get supported datatypes
      *
+     * @psalm-suppress MixedReturnStatement
```
gsteel: Please can you drop this inline suppression?

It's better off in the baseline, and it'll mean fewer conflicts when merging up into 4.x

test/Storage/Adapter/AbstractAdapterTest.php
```diff
@@ -497,7 +497,7 @@ public function testEventHandlingStopInPre(
         $eventManager->attach($methodName . '.post', $eventHandler);
         $eventManager->attach($methodName . '.exception', $eventHandler);
 
-        $eventManager->attach($methodName . '.pre', static function ($event) use ($retVal) {
+        $eventManager->attach($methodName . '.pre', static function ($event) use ($retVal): mixed {
```
gsteel: If Psalm is forcing this return type addition, can we baseline this too please? Purely to reduce merge-up conflicts.

gsteel (commented):
Thanks @driehle - If we can keep the diff just to dot files and tooling stuff, the merge-up will be cleaner 🤞

Also, yes to leaving the docs linting errors alone 👍

src/Storage/Capabilities.php
driehle: Done with e099a96

test/Storage/Adapter/AbstractAdapterTest.php
driehle: Done with e099a96

gsteel (approved):
Thank you @driehle 👍

driehle:
@gsteel @froschdesign 

As a follow-up on this PR and the release of laminas-cache 3.13.0 (thanks, @gsteel!), I have created the following PRs to add support for PHP 8.4 to the relevant storage adapters:

- laminas/laminas-cache-storage-adapter-blackhole#66
- laminas/laminas-cache-storage-adapter-filesystem#103
- laminas/laminas-cache-storage-adapter-memory#69
- laminas/laminas-cache-storage-adapter-redis#135
- laminas/laminas-cache-storage-adapter-apcu#57
- laminas/laminas-cache-storage-adapter-memcached#75

I would greatly appreciate, if you could have a look at them to get them merged and a release tagged. Thanks!

driehle:
I missed that we need laminas-serializer as well to get laminas-cache running with PHP 8.4. So there are two more PRs on laminas-serializer 2.x now, of which the first we'll need to merge as well. The second is rather just a suggestion.

- laminas/laminas-serializer#73
- laminas/laminas-serializer#74

Issue #355, opened 2025-01-08
ObjectCache fails to create consistent cache keys in V4


SvenRtbg:
Side remark: There seems to be no CHANGELOG.md that I could consult prior to reporting the BC break

### BC Break Report

The `\Laminas\Cache\Pattern\ObjectCache` class changed the way it creates the cache key between V3 and v4.0.0.

|    Q        |   A
|------------ | ------
| Version     | 4.0.0 up to 4.1.0 

#### Summary

V4.0.0 and up broke apart the inheritance chain of `ObjectCache` and `CallbackCache`. This introduced a significant behaviour change when using `ObjectCache` which may make it impossible to create cache hits anymore.

#### Previous behavior

At the end of executing `ObjectCache::call`, the call is passed to `CallbackCache` as `parent::call()`, and there the line `$key     = $this->generateCallbackKey($callback, $args);` does jump back to `ObjectCache::generateCallbackKey()`, which returns the name of the cached object class and the method as a string.

#### Current behavior

Starting with V4.0.0, there is no inheritance chain, and above mentioned line jumps to `CallbackCache::generateCallbackKey()`, which does quite some logic.

The most obvious difference is that now the object that is to be cached is passed to `\serialize($object)` to create the cache key. This is bad for multiple reasons, as the object tree that is actually to be cached might be quite large.

But the obvious problem we encountered is that inside this object tree that is supposed to execute a SOAP call, we have a profiler class that stores the current time. Said time stamp now is part of the cache key, and will change, effectively disabling the cache.

As a side effect, I discovered that the logic creating the callback key explicitly mentions `Closure` as a case in the comment, but does not prevent that `Closure` to be serialized later - which PHP still cannot do. Again, closures embedded in the object tree will prevent caching as they cannot be serialized with `CallbackCache`, compared to the previous code in `ObjectCache`.

#### How to reproduce

I'll try to create a test case with details.

In summary, any object stack that contains dynamic data in properties will yield changing cache keys.

Injecting this class to be cached may work when executed in the same runtime context, but will fail when used in a web server context.
```php
class Foo { 
    private int $time; 
    public function __construct() 
    { 
        $this->time = time();
    }
    public function cachedMethod(): string
    {
        return "some_cacheable_value";
    }
}
```

If a closure is present in any property, serialization will fail.

froschdesign:
Please check the documentation https://docs.laminas.dev/laminas-cache/v4/migration/to-version-4/
Or the release notes.

[...]

PR #358, opened 2025-01-14
Fix proposal 355

SvenRtbg:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

I discovered that `\Laminas\Cache\Pattern\ObjectCache`, after being untangled and not inheriting from `CallbackCache` anymore, has changed the way the cache key computation is done and now is sensitive to all properties being present, and cannot deal with Closures present in them.

This is described in issue #355.

Expected behaviour (identical with V3.x):
- `ObjectCache` allows to cache object trees that contain Closures, as they are not used to create a cache key.
- `ObjectCache` is consistently creating the same cache key independent from any property values that are contained in the object tree.

Observed behavior with V4.0 up to v4.1.0:
- `ObjectCache` throws an exception when encountering a Closure in a property.
- `ObjectCache` creates differing cache keys if properties change inside the object tree.

This proposal adds a test case indicator by simply forcing a useless Closure into the `\LaminasTest\Cache\Pattern\TestAsset\TestObjectCache`, which prevents serializing this object.

Then a code path is added to `CallbackCache` that detects if an object is present in the `PatternOptions`, which is taken as an indicator we are dealing with `ObjectCache` forwarding the call into `CallbackCache`, and apply the previously existing cache key logic that was present in `ObjectCache::generateCallbackKey()` from version 3.x. This will restore the previous behavior.

This PR targets the 4.2.x branch, as there is currently only a minor lock file commit after the 4.1.x branch, and merging this fix may be considered serious enough to warrant a new minor release, as it reverts a major version change I consider to be accidental.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -394,6 +394,9 @@
     <PossiblyUnusedProperty>
       <code><![CDATA[$property]]></code>
     </PossiblyUnusedProperty>
+    <UnusedProperty>
+      <code><![CDATA[$closure]]></code>
+    </UnusedProperty>
   </file>
   <file src="test/Psr/CacheItemPool/CacheItemPoolDecoratorTest.php">
     <InvalidArgument>
--- a/src/Pattern/CallbackCache.php
+++ b/src/Pattern/CallbackCache.php
@@ -7,6 +7,7 @@ use Throwable;
 
 use function array_key_exists;
 use function array_values;
+use function is_array;
 use function is_callable;
 use function is_object;
 use function md5;
@@ -111,17 +112,16 @@ final class CallbackCache extends AbstractStorageCapablePattern
      */
     protected function generateCallbackKey(callable $callback, array $args): string
     {
+        if (! is_callable($callback, false, $callbackKey)) {
+            throw new Exception\InvalidArgumentException('Invalid callback');
+        }
 
         // Create a cache key for the ObjectCache use case.
         $options = $this->getOptions();
-        if (is_object($options->getObject())) {
+        if (is_object($options->getObject()) && is_array($callback) && isset($callback[1])) {
             $callbackKey = md5($options->getObjectKey() . '::' . strtolower($callback[1]));
             $argumentKey = $this->generateArgumentsKey($args);
             return $callbackKey . $argumentKey;
-        }
-
-        if (! is_callable($callback, false, $callbackKey)) {
-            throw new Exception\InvalidArgumentException('Invalid callback');
         }
 
         // functions, methods and classnames are case-insensitive
--- a/src/Pattern/CallbackCache.php
+++ b/src/Pattern/CallbackCache.php
@@ -111,6 +111,15 @@ final class CallbackCache extends AbstractStorageCapablePattern
      */
     protected function generateCallbackKey(callable $callback, array $args): string
     {
+
+        // Create a cache key for the ObjectCache use case.
+        $options = $this->getOptions();
+        if (is_object($options->getObject())) {
+            $callbackKey = md5($options->getObjectKey() . '::' . strtolower($callback[1]));
+            $argumentKey = $this->generateArgumentsKey($args);
+            return $callbackKey . $argumentKey;
+        }
+
         if (! is_callable($callback, false, $callbackKey)) {
             throw new Exception\InvalidArgumentException('Invalid callback');
         }
--- a/test/Pattern/TestAsset/TestObjectCache.php
+++ b/test/Pattern/TestAsset/TestObjectCache.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace LaminasTest\Cache\Pattern\TestAsset;
 
+use Closure;
+
 use function func_get_args;
 use function implode;
 
@@ -22,13 +24,15 @@ final class TestObjectCache
     /** @var string */
     public $property = 'testProperty';
 
-    private \Closure $closure;
+    private Closure $closure;
 
     public function __construct()
     {
         // Closures prevent serialization - this acts as a detector to verify this object is not serialized during test.
-        $this->closure = function () {};
+        $this->closure = function (): void {
+        };
     }
+
     public function bar(): string
     {
         ++static::$fooCounter;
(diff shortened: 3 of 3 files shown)


froschdesign:
@boesing 
Is this correct?

SvenRtbg:
I'd like to draw attention towards this PR again, as I would still like it to be merged.

froschdesign:
@boesing 
If you could find some time, I would be grateful if you could let us know what you think of this correction.
Thanks in advance! 👍🏻

boesing:
Seems that there was no test for this before. Thanks for adding this!

However, I am not really happy with the fact that the `CallbackCache` now has details about certain object-cache specifics. If we now add this to the callback cache, it will be most-likely impossible to remove this in the future.

Another problem is, that this now only supports closure serialization for `ObjectCache` while `CallbackCache` can still not be used with closures. There is no test which verifies that closures in general work, such as:

```php
$class = new class {
    public function callback(): void {}
};

$callbackCache = new CallbackCache();
$callbackCache->generateKey($class->callback(...), []);
```

```php
$callbackCache = new CallbackCache();
$callbackCache->generateKey(fn () => null, []);
```

I get the point that this PR is fixing an issue which worked before. But that was mainly due to the inheritance which was properly "replacing" the `CallbackCache#generateKey` method. Now, we have to add details to the `CallbackCache` that it is also used by the `ObjectCache`. IMHO thats some kind of design flaw and does not really solve the underlying issue with unserializable callbacks in general (such as closures).

I wonder if it might make sense to use a different approach than `serialize` the object. Do you know why the serialization might be necessary? I have never used these cache patterns and thus I can't wrap my head around the fact why an object state should be used inside a cache key in the first place. A bit more details would really help to properly add support for `Closure` in general. As of today, closures are not supported at all when it comes to `CallbackCache`, neither by having a closure in an `ObjectCache` `object` nor when it comes to passing callables as closures via the new `$object->methodName(...)` syntax.

---

In addition to this, I came accross something like https://github.com/opis/closure 
Maybe this should be used for serializing the callback rather than using native PHP `serialize`. But again, I wonder if it even makes sense to serialize objects in the first place as objects might have a state which might lead to different cache keys depending on what the object is actually containing...

SvenRtbg:
I'd argue that the current `CallbackCache` already has knowledge about the case when an object is passed inside the callback array structure. It isn't exactly what you are referring to, as the object in the array structure counts as a genuine callback of a single method, but I felt this approach somewhat supports my strategy to fix the issue. There's basically no alternative given that every class is made final, no extension points are available, and `ObjectCache` heavily utilizes `CallbackCache` internally - basically the only alternative would be code duplication, or maybe code extraction into some collection of traits.

Re `CallbackCache` not supporting closures: I haven't checked, but would believe this never worked previously, as there is a need to identify the fingerprint of the code to be executed, i.e. the name of the function/method that is to be called. Closures don't have a name, their fingerprint basically is the code or bytecode they consist of, which seems to be inaccessible by pure PHP means - the opis/closure approach seems to search for the source code file and reads the code from disk, which sounds not exactly like it's fast - haven't measured anything, though.

In essence: The 3.x versions used inheritance to put code for the exact usecase in place - which explains why `CallbackCache::generateKey()` still directly forwards to the  `protected` method `CallbackCache::generateCallbackKey()` - it seems useless in a `final` class, but illustrates the former extension point.

boesing:
I at least tested with opis/closure and replaced the serialize function with the one from that library. That fixed all use-cases, including yours plus all the cases I am refering to in the other comment. 

From what I can say, the CallbackCache has no specific knowledge about the ObjectCache, yet it will have when merging this.

I am not 100% certain if it makes sense to actually add a new dependency, yet I wonder why it is sufficient to use the object key (as it seems to be a class-string unless a specific one is configured).

Do you actually provide a dedicated object key via options or do you rely on the class-string only? 

---

However, I would slightly refactor the code provided here to make use of the is_array check. 

I was already coding on this and had something like:

```php
        $object = $callback;

        // generate a unique key of object callbacks
        if (is_array($callback)) {
            // array($object, 'method')
            $object = $callback[0];
            
            $options = $this->getOptions();
            if ($options->getObject() === $object) {
                $callbackKey = md5($options->getObjectKey() . '::' . strtolower($callback[1]));
                $argumentKey = $this->generateArgumentsKey($args);
                return $callbackKey . $argumentKey;
            }
        }
```

If you rebase this PR against 4.4.x, psalm will not argue about `$callback[1]` anymore due to enhanced types in psalm v6.

Could you refactor the code in this direction? I'd say lets merge it then - yet this still does not fix other `Closure` cases.
But you are right, this never worked, yet the method allows any `callable` which includes the examples above.

---

IMHO, the whole feature should be removed at some point anyway. I do not really see value in it on how it works like now. I have never understood a real-world use-case (due to lack of understanding on how this would be used).

If you could provide some real-world example, I could wrap my head around a proper replacement which leverages a better UI than what we have right now. I guess, in the end, this should somehow provide a cache wrapper for certain ojects, yet it kind-of loses interface capabilities. Maybe a dedicated component, which uses ocramius/proxy-manager or sth. to implement an interface of the object passed to it so that one can do from what the [example code for filters](https://docs.laminas.dev/laminas-cache/v4/pattern/object-cache/#caching-a-filter) says:

```php
use Laminas\Cache\Pattern\ObjectCache;
use Laminas\Cache\Pattern\PatternOptions;

final readonly class Whatever
{
    public function __construct(private \Laminas\Filter\FilterInterface $filter) {}
}

$filter  = new \Laminas\Filter\RealPath();
// Some magic to build a cached filter which is actually implementing `FilterInterface`
$cachedFilter = null;
new Whatever($cachedFilter);
```

IMHO, that would provide way better re-usability, especially since there is a decorator in place which can be configured.
That would work way better than what we have right now TBH. But I do not really see that we are implementing this on our own. Again, I get the point that we have to support it as it is a regression - yet I was about to remove that feature when working on v4 as I was not expecting anyone is using this (as it is not really re-usable, thats why I am curious about real-world examples as I barely do not understand how this would look like in actual production code).

Thus said, once the refactoring is done, I guess we are good to go.
Thanks for pointing out the regression.

PR #378, opened 2026-01-30
Allow PHP 8.5, remove PHP 8.1

### LINKED ISSUE #377: PHP 8.5 support
The current PR #374 is blocked by no 8.5 support in laminas/laminas-serializer. See https://github.com/laminas/laminas-serializer/issues/84

KorvinSzanto:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes
| BC Break      | no

### Description
 - Remove deprecated tests functionality
 - Fix psalm issues
 - Set exceptions final
 - Add `@final` soft-final to everything else Psalm complains about

See #377 and https://github.com/laminas/laminas-serializer/pull/85

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "process-timeout": 600,
         "sort-packages": true,
         "platform": {
-            "php": "8.1.99"
+            "php": "8.2.99"
         },
         "allow-plugins": {
             "dealerdirect/phpcodesniffer-composer-installer": true
@@ -34,9 +34,9 @@
         }
     },
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
+        "php": "~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0",
         "laminas/laminas-eventmanager": "^3.4",
-        "laminas/laminas-servicemanager": "^4.1",
+        "laminas/laminas-servicemanager": "^4.5",
         "laminas/laminas-stdlib": "^3.20",
         "psr/cache": "^2.0 || ^3.0",
         "psr/clock": "^1.0",
@@ -47,10 +47,10 @@
         "laminas/laminas-cli": "^1.11",
         "laminas/laminas-coding-standard": "~3.1.0",
         "laminas/laminas-config-aggregator": "^1.17",
-        "laminas/laminas-serializer": "^3.1",
-        "phpunit/phpunit": "^10.5.38",
+        "laminas/laminas-serializer": "^3.2",
+        "phpunit/phpunit": "^11.5.37",
         "psalm/plugin-phpunit": "^0.19.0",
-        "vimeo/psalm": "^5.26.1"
+        "vimeo/psalm": "^6"
     },
     "conflict": {
         "laminas/laminas-serializer": "<3.0",
--- a/src/Storage/Adapter/AdapterOptions.php
+++ b/src/Storage/Adapter/AdapterOptions.php
@@ -170,6 +170,8 @@ class AdapterOptions extends AbstractOptions
 
     /**
      * Set time to live.
+     *
+     * @param int|float|numeric-string $ttl
      */
     public function setTtl(int|float|string $ttl): self
     {
--- a/src/Storage/Adapter/AdapterOptions.php
+++ b/src/Storage/Adapter/AdapterOptions.php
@@ -15,6 +15,7 @@ use Webmozart\Assert\Assert;
 use function array_change_key_case;
 use function array_reverse;
 use function array_shift;
+use function assert;
 use function is_array;
 use function is_int;
 use function iterator_to_array;
@@ -169,8 +170,6 @@ class AdapterOptions extends AbstractOptions
 
     /**
      * Set time to live.
-     *
-     * @param numeric $ttl
      */
     public function setTtl(int|float|string $ttl): self
     {
@@ -227,7 +226,6 @@ class AdapterOptions extends AbstractOptions
     /**
      * Validates and normalize a TTL.
      *
-     * @param numeric $ttl
      * @return non-negative-int|float $ttl
      * @throws Exception\InvalidArgumentException
      */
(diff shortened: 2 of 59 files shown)


KorvinSzanto:
@froschdesign would you be the one to review / merge? A couple things that could change are:
- I could just baseline all the pest concerns rather than try to fix the low hanging ones
- I could soft-final the exceptions to preserve backward compatibility

froschdesign:
@KorvinSzanto 
> * I could soft-final the exceptions to preserve backward compatibility

You can trust the checks, so please do this.
Thanks in advance! 👍🏻

KorvinSzanto:
Alright @froschdesign should be good to review

composer.json
```diff
@@ -34,9 +34,9 @@
         }
     },
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
+        "php": "~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0",
         "laminas/laminas-eventmanager": "^3.4",
-        "laminas/laminas-servicemanager": "^4.1",
+        "laminas/laminas-servicemanager": "^4.5",
```
gsteel, line 39: Is there a specific reason why this needed a bump?

This patch will go into a minor, so the dependency change is OK, but if it's just for minimum version supporting 8.5 - it's not necessary because composer takes care of this

src/Storage/Adapter/AdapterOptions.php
```diff
@@ -169,8 +170,6 @@ public function getReadable(): bool
 
     /**
      * Set time to live.
-     *
-     * @param numeric $ttl
```
gsteel, line 173: ideally, this would be changed to `int|float|numeric-string` rather than being dropped - it's good information for static analysers

src/Storage/IterableInterface.php
```diff
@@ -3,13 +3,19 @@
...
  * @template-extends IteratorAggregate<TKey, TValue>
  */
 interface IterableInterface extends IteratorAggregate
 {
+    /**
+     * Get the storage iterator
+     *
+     * @return Traversable<TKey, TValue>|TValue[]
+     */
+    public function getIterator();
```
gsteel: I don't understand why the method is added here - I agree with dropping the `@method`, but `getIterator(): Traversable<TKey, TValue>` is already present on `IteratorAggregate`

test/Psr/CacheItemPool/CacheItemTest.php
```diff
@@ -107,8 +108,10 @@ public function testExpiresAfterNull(): void
 
     public function testExpiresAfterStartsExpiringAfterMethodCall(): void
     {
-        $now              = new DateTimeImmutable();
-        $nowPlusOneSecond = $now->add(DateInterval::createFromDateString('1 second'));
+        $now      = new DateTimeImmutable();
+        $interval = DateInterval::createFromDateString('1 second');
+        assert($interval instanceof DateInterval);
+        $nowPlusOneSecond = $now->add($interval);
```
gsteel: Nit pick, this could be reduced to `$whatever = (new DateTimeImmutable())->add(new DateInterval('PT1S'));` - feel free to ignore

test/Storage/TestAsset/ClearExpiredMockAdapter.php
```diff
@@ -10,6 +10,7 @@
 /**
  * @template TOptions of AdapterOptions
  * @template-extends MockAdapter<TOptions>
+ * @final
  */
 class ClearExpiredMockAdapter extends MockAdapter implements ClearExpiredInterface
```
gsteel, line 15:
```suggestion
 */
final class ClearExpiredMockAdapter extends MockAdapter implements ClearExpiredInterface
```

test/Storage/TestAsset/MockPlugin.php
```diff
@@ -9,6 +9,7 @@
 use Laminas\EventManager\Event;
 use Laminas\EventManager\EventManagerInterface;
 
+/** @final */
 class MockPlugin extends AbstractPlugin
```
gsteel:
```suggestion
final class MockPlugin extends AbstractPlugin
```

test/Storage/TestAsset/OptimizableMockAdapter.php
```diff
@@ -10,6 +10,7 @@
 /**
  * @template TOptions of AdapterOptions
  * @template-extends MockAdapter<TOptions>
+ * @final
  */
 class OptimizableMockAdapter extends MockAdapter implements OptimizableInterface
```
gsteel, line 15:
```suggestion
 */
final class OptimizableMockAdapter extends MockAdapter implements OptimizableInterface
```

test/Storage/PluginManagerTest.php
```diff
@@ -11,6 +11,7 @@
 use Laminas\ServiceManager\Test\CommonPluginManagerTrait;
 use PHPUnit\Framework\TestCase;
 
+/** @final */
 class PluginManagerTest extends TestCase
```
gsteel: Same with anything in `/test/` - introducing the `final` keyword is fine 👍
```suggestion
final class PluginManagerTest extends TestCase
```

gsteel (changes requested):
Thanks very much @KorvinSzanto - a few minor points is all.

Sorry it took a while to review

composer.json
KorvinSzanto: I believe this was done due to psalm complaints but I don't recall exactly. I'd leave it as is given that `laminas/serializer` [requires 4.5](https://github.com/laminas/laminas-serializer/blob/3.3.x/composer.json#L36) and prevents us from testing with lower versions of servicemanager

src/Storage/IterableInterface.php
KorvinSzanto: Originally this was added to declare the return type due to a psalm complaint, but I agree this can go away now.

test/Storage/TestAsset/OptimizableMockAdapter.php
KorvinSzanto: This one has to be soft to allow mocking

test/Storage/TestAsset/ClearExpiredMockAdapter.php
KorvinSzanto: This one has to be soft to allow mocking

gsteel (approved):
Thanks again @KorvinSzanto

PR #380, opened 2026-02-04
3.x: Remove support for PHP 8.1, add support for PHP 8.5

KorvinSzanto:
In the same spirit of https://github.com/laminas/laminas-servicemanager/pull/267 3.x should support PHP 8.5

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -6,9 +6,5 @@
         "apc.enabled=1",
         "apc.enable_cli=1"
     ],
-    "backwardCompatibilityCheck": true,
-    "ignore_php_platform_requirements": {
-        "8.4": true,
-        "8.5": true
-    }
+    "backwardCompatibilityCheck": true
 }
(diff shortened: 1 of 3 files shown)


KorvinSzanto:
@gsteel I believe this is ready to review, the failing check is also failing in the upstream branch.

.laminas-ci.json
```diff
@@ -8,6 +8,7 @@
     ],
     "backwardCompatibilityCheck": true,
     "ignore_php_platform_requirements": {
-        "8.4": true
+        "8.4": true,
```
froschdesign, line 11: This entry should no longer be necessary, so please remove it and see if there are any problems.
Thanks in advance! 👍🏻

KorvinSzanto: Looks like it's still necessary

gsteel (approved):
Thanks @KorvinSzanto

PR #388, opened 2026-03-10
Allow webmozart/assert:^2.1.6

fezfez:


gsteel (approved):
Thanks @fezfez

src/Service/StorageAdapterFactory.php
```diff
@@ -71,7 +71,8 @@ public function create(string $storage, array $options = [], array $plugins = []
     public function assertValidConfigurationStructure(array $configuration): void
     {
         try {
-            Assert::isNonEmptyMap($configuration, 'Configuration must be a non-empty array.');
+            Assert::isMap($configuration, 'Configuration must be a non-empty array.');
```
samsonasik, line 74: `isNonEmptyMap()` exists in both v1 and v2

https://github.com/webmozarts/assert/blob/9be6926d8b485f55b9229203f962b51ed377ba68/src/Assert.php#L2148

https://github.com/webmozarts/assert/blob/ff31ad6efc62e66e518fbab1cde3453d389bcdc8/src/Assert.php#L2115

I guess this change is not needed?

fezfez: psalm complain about this
```
Error: src/Service/StorageAdapterFactory.php:74:35: MixedArgumentTypeCoercion: Argument 1 of Webmozart\Assert\Assert::isNonEmptyMap expects array<string, mixed>, but parent type array<array-key, mixed> provided (see https://psalm.dev/194)
Error: src/Service/StoragePluginFactory.php:39:35: MixedArgumentTypeCoercion: Argument 1 of Webmozart\Assert\Assert::isNonEmptyMap expects array<string, mixed>, but parent type array<array-key, mixed> provided (see https://psalm.dev/194)
```

gsteel: Doc blocks/types/assertions have regressed somewhat in v2 afair. Merging as-is.

